PR #4, opened 2019-11-07
Entry Point Test

jeff-hernandez:
This adds an entry point test to the integration tests.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,7 @@
 numpy>=1.13.3
-pandas>=0.23.0
+pandas>=0.23.0,<0.24.0
 tqdm>=4.19.2
-featuretools>=0.7.0
+featuretools>=0.7.0,<0.12.0
 nltk>=3.4
 tensorflow<=1.14.0
 tensorflow_hub<=0.4.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,7 @@
 numpy>=1.13.3
-pandas>=0.23.0,<0.24.0
+pandas>=0.23.0
 tqdm>=4.19.2
-featuretools>=0.7.0,<0.12.0
+featuretools>=0.7.0
 nltk>=3.4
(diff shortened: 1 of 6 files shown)


requirements.txt
```diff
@@ -1,9 +1,10 @@
 numpy>=1.13.3
-pandas>=0.23.0
+pandas>=0.23.0,<0.24.0
```
rwedge: Does this library need to restrict itself to pandas 0.23?

requirements.txt
```diff
@@ -1,9 +1,10 @@
...
-featuretools>=0.7.0
+featuretools>=0.7.0,<0.12.0
 nltk>=3.4
 tensorflow<=1.14.0
 tensorflow_hub<=0.4.0
 scikit-learn>=0.20.0,<0.21; python_version < '3'
-scikit-learn>=0.20.0; python_version >= '3'
\ No newline at end of file
+scikit-learn>=0.20.0; python_version >= '3'
+python-dateutil==2.8.0
```
rwedge: I think a range would be better, for featuretools I used
`python-dateutil>=2.6.1,<2.8.1`

requirements.txt
```diff
@@ -1,9 +1,10 @@
 numpy>=1.13.3
-pandas>=0.23.0
+pandas>=0.23.0,<0.24.0
 tqdm>=4.19.2
-featuretools>=0.7.0
+featuretools>=0.7.0,<0.12.0
```
rwedge: Same question with restricting featuretools

requirements.txt
jeff-hernandez: not necessary, reverted back to 0.23

requirements.txt
jeff-hernandez: updated to range

requirements.txt
jeff-hernandez: reverted back to >=0.7.0

PR #7, opened 2019-11-15
Requirement Updates

jeff-hernandez:
This updates requirements to later versions.


requirements.txt
```diff
@@ -3,8 +3,8 @@ pandas>=0.23.0
 tqdm>=4.19.2
```
rwedge: Do we use tqdm directly in this library?

jeff-hernandez: It doesn't seem we do. Should I remove?

rwedge: Yeah let's remove

jeff-hernandez: removed

PR #8, opened 2019-11-15
v0.2.4

jeff-hernandez:
**v02.4** November 15, 2019
  * Added Featuretools Entry Point (#4)
  * PyPI Upload (#5)

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -3,7 +3,7 @@ Changelog
 =========
 
 **v0.2.4**
-    * Entry Point Test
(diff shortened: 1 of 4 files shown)


docs/source/changelog.rst
```diff
@@ -2,6 +2,10 @@
 Changelog
 =========
 
+**v0.2.4**
+    * Entry Point Test
```
rwedge: Can we change this to "Added Featuretools Entry Point"

jeff-hernandez: updated

PR #11, opened 2020-04-20
Update python-dateutil version

gsheni:
- installing `featuretools[complete]` shows the following error:
```
ERROR: nlp-primitives 0.2.4 has requirement python-dateutil<2.8.1,>=2.6.1, but you'll have python-dateutil 2.8.1 which is incompatible.
```
- This attempts to resolve this by increasing the version required (matches the python version)


rwedge:
Let's try removing python-dateutil as a requirement, it doesn't look like we import it directly.

There was an issue with installing the latest python-dateutil version back in November, it should be resolved now.

gsheni:
> Let's try removing python-dateutil as a requirement, it doesn't look like we import it directly.
> 
> There was an issue with installing the latest python-dateutil version back in November, it should be resolved now.

Fixed.

PR #12, opened 2020-04-20
v0.2.5

gsheni:
v0.2.5 April 20, 2020
- Removed python-dateutil as a requirement

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -2,7 +2,7 @@
 Changelog
 =========
 
-**v0.2.4**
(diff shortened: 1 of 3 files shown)


docs/source/changelog.rst
```diff
@@ -2,6 +2,9 @@
 Changelog
 =========
 
+**v0.2.4**
```
rwedge, line 5: This should be 0.2.5

gsheni: Fixed

PR #14, opened 2020-05-14
Simplify Package Setup

jeff-hernandez:
The package was simplified by the following changes:
-  consolidating dev requirements into test requirements
- removing an unnecessary entry point reference
- removing the progress bar during package installations for integration tests


rwedge:
Can we also update the test requirement package versions?

rwedge:
we can also remove codecov, pytest-cov, and pytest-xdist since they are unused

PR #16, opened 2020-05-15
Update Python Versions

jeff-hernandez:
Dropped support for Python 3.5 and added Python 3.8.


rwedge:
We should also remove the python 2 specific requirements in requirements.txt

PR #21, opened 2020-07-29
Fit LSA primitive on given text rather than downloaded corpus

eccabay:
Should fix evalml issues [#940](https://github.com/featurelabs/evalml/issues/940) and [#980](https://github.com/featurelabs/evalml/issues/980) upon release.

### DIFF
--- a/nlp_primitives/lsa.py
+++ b/nlp_primitives/lsa.py
@@ -19,7 +19,7 @@ class LSA(TransformPrimitive):
         value decomposition to go from a sparse matrix to a compact matrix with two
         values for each string. These values represent that Latent Semantic Analysis
         of each string. These values will represent their context with respect to
-        (nltk's brown sentence corpus.)[https://www.nltk.org/book/ch02.html#brown-corpus]
+        the corpus of all strings in the given list.
 
         If a string is missing, return `NaN`.
 
@@ -29,18 +29,16 @@ class LSA(TransformPrimitive):
         >>> res = lsa(x).tolist()
         >>> for i in range(len(res)): res[i] = [abs(round(x, 2)) for x in res[i]]
         >>> res
-        [[0.0, 0.0, 0.01], [0.0, 0.0, 0.0]]
+        [[0.0, 0.0, 1.0], [0.0, 1.0, 0.0]]
 
-        Now, if we change the values of the input corpus, to something that better resembles
-        the given text, the same given input text will result in a different, more discerning,
-        output. Also, NaN values are handled, as well as strings without words.
+        NaN values are handled, as well as strings without words.
 
         >>> lsa = LSA()
-        >>> x = ["the earth is round", "", np.NaN, ".,/"]
+        >>> x = ["the earth is round", "", np.NaN, ".,/", "the sun is a star"]
         >>> res = lsa(x).tolist()
         >>> for i in range(len(res)): res[i] = [abs(round(x, 2)) for x in res[i]]
         >>> res
-        [[0.01, 0.0, nan, 0.0], [0.0, 0.0, nan, 0.0]]
+        [[1.0, 0.0, nan, 0.0, 0.0], [0.0, 0.0, nan, 0.0, 1.0]]
 
     """
     name = "lsa"
@@ -49,18 +47,10 @@ class LSA(TransformPrimitive):
     default_value = 0
 
     def __init__(self):
-        # TODO: allow user to use own corpus
         self.number_output_features = 2
         self.n = 2
 
-        try:
-            brown = nltk.corpus.brown.sents()
-        except LookupError:
-            nltk.download('brown')
-            brown = nltk.corpus.brown.sents()
-        finally:
-            self.trainer = make_pipeline(TfidfVectorizer(), TruncatedSVD())
-            self.trainer.fit([" ".join(sent) for sent in brown])
+        self.trainer = make_pipeline(TfidfVectorizer(), TruncatedSVD(random_state=42))
 
     def get_function(self):
         dtk = TreebankWordDetokenizer()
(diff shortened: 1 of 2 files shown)


nlp_primitives/lsa.py
```diff
@@ -49,18 +47,10 @@ class LSA(TransformPrimitive):
...
 
-        try:
-            brown = nltk.corpus.brown.sents()
-        except LookupError:
-            nltk.download('brown')
-            brown = nltk.corpus.brown.sents()
-        finally:
-            self.trainer = make_pipeline(TfidfVectorizer(), TruncatedSVD())
-            self.trainer.fit([" ".join(sent) for sent in brown])
+        self.trainer = make_pipeline(TfidfVectorizer(), TruncatedSVD(random_state=42))
```
ctduffy, line 53: Might it be worth allowing the users to input their own random state as an optional argument with a default of 42?

nlp_primitives/tests/test_lsa.py
```diff
@@ -15,27 +15,21 @@ def test_strings(self):
...
-                                             decimal=2)
+        answers = np.array([[0.7751553, 0.7751553, 0., 0.], [0., 0., 1., 0.]])
+        results = np.array(primitive_func(x).values.tolist())
+        np.testing.assert_array_almost_equal(answers, results)
 
     def test_nan(self):
         x = pd.Series([np.nan,
                        '#;.<',
                        'This IS a STRING.'])
         primitive_func = self.primitive().get_function()
```
ctduffy, line 26: I think it would also be useful to see a test of what happens when a list with one string (maybe one testing the nan case, and another testing a regular string) is passed in. Something like `x = pd.Series([np.nan])` and `x = pd.Series(['string'])`.

nlp_primitives/lsa.py
```diff
@@ -69,6 +60,13 @@ def lsa(array):
             array = pd.Series(array, index=pd.Series(array.index), name='array')
             copy = array.dropna()
             copy = copy.apply(lambda x: dtk.detokenize(clean_tokens(x)))
+
+            fit_data = copy.tolist()
+            # TruncatedSVD cannot produce two features without multiple input values
+            if len(fit_data) == 1:
+                fit_data = fit_data * 2
```
dsherry, line 67: So this case is when only one row has been passed in as input, right? If so, perhaps we should just raise an exception. And then featuretools and evalml can both handle that exception.

nlp_primitives/lsa.py
```diff
@@ -69,6 +60,13 @@ def lsa(array):
             array = pd.Series(array, index=pd.Series(array.index), name='array')
             copy = array.dropna()
             copy = copy.apply(lambda x: dtk.detokenize(clean_tokens(x)))
+
+            fit_data = copy.tolist()
```
dsherry, line 64: Oh weird, `TfidfVectorizer` can't handle `pandas` or `numpy` types?

nlp_primitives/lsa.py
```diff
@@ -29,38 +29,29 @@ class LSA(TransformPrimitive):
...
 
     """
     name = "lsa"
     input_types = [Text]
     return_type = Numeric
     default_value = 0
 
-    def __init__(self):
-        # TODO: allow user to use own corpus
+    def __init__(self, random_state=42):
```
dsherry, line 49: @rwedge @ctduffy @eccabay : is there a way to provide parameters to primitives in featuretools? It would be good to know how hard it is to change the value of `random_state` here.

I also ask because we could get another PR up after this which adds the ability to provide a different corpus to use for TF-IDF, to preserve the existing functionality of using the `nltk` "brown" corpus. @rwedge do you think its a requirement for featuretools that we preserve that functionality before doing another release of `nlp_primitives`?

rwedge: @dsherry it is pretty simple to provide parameter by passing a instantiated primitive to DFS instead of a string or the class

eg setting the dfs parameter `trans_primitives` to `trans_primitives=[LSA(random_seed=27)]` instead of `trans_primitives=["lsa"]`

dsherry: Awesome, and then we could do the same in evalml [in the text featurizer](https://github.com/FeatureLabs/evalml/blob/main/evalml/pipelines/components/transformers/preprocessing/text_featurizer.py#L85). Nice.

So what about this, in the lsa primitive?
`def __init__(self, custom_training_data=None, random_state=42):`

Where we do the following with the `custom_training_data` arg:
* If `None` (default), use existing code to download the `nltk` "brown" corpus at runtime for TF-IDF
* If `'training_data'`, do what @eccabay 's PR currently does, and use the training data for TF-IDF
* Else if a pd dataframe (or should this be np array, or list? all of the above), store that corpus on the primitive and use it for TF-IDF

What do you all think of that? It would preserve the existing functionality by default, and it would allow us in evalml to specify whatever training data we'd like to try.

rwedge: One thing with the current approach is that treating the data passed to the `lsa` function as the training data is that it could be highly compartmentalized piece of the general dataset.  

Say we had a customers table with an "about" text field, and wanted to create an LSA(about) feature.

If we decide to calculate the feature values for all the customers using all the data, the entire "about" column will be passed in to the `lsa` function and used to fit the pipeline.

If instead we calculate the feature values for two customers, at two different points in time, we will call the `lsa` function twice - once with all the data for customer 1 before a specific time, and once with all the data for customer 2 before customer 2's specific time

dsherry: Ah, thanks for pointing that out @rwedge . I didn't think through what data would be provided to this primitive by featuretools.

From evalml's perspective, I think our goal is to be able to fit TF-IDF on a corpus drawn from the entire training data provided to evalml's `search` method.

A question which comes to mind here: should that corpus consist of the data from one particular text column, or should it be the union of all the text columns? I think the first thing we should try is the union of all text in the training data.

Another consideration: it would be ideal to avoid having a raw reference to the data hang around in evalml after the training has completed and TF-IDF has been fit. Because at that point we don't need that reference, and it could be large.

Let's do the following:
* Close this PR instead of merging, since we have some more details to resolve before this is ready to go.
* @eccabay talk to @rwedge (or someone from featuretools) about how to do this.
* Post the new plan on the issue and let's discuss it further. Then once we resolve that discussion we can get another PR up.

That work for you all?

PR #23, opened 2020-07-30
Remove LSA online dependency

eccabay:
As per the discussion [here](https://github.com/FeatureLabs/nlp_primitives/pull/21), the previous behavior of the LSA primitive to use a pipeline fitted on nltk's brown corpus is maintained, but there is now the ability to pass in a pre-fitted trainer object to the primitive instead.

The decision to require the trainer to be pre-fitted stems from the requirement of all featuretools primitives to have every parameter passed into the primitive's constructor be saved as a class attribute. If a corpus was passed in here, it would be saved unnecessarily within the primitive. Therefore, while there is now a function to train a LSA pipeline given a corpus, it does not happen in the init function so as to keep memory requirements lower.

### DIFF
--- a/nlp_primitives/lsa.py
+++ b/nlp_primitives/lsa.py
@@ -6,7 +6,7 @@ from featuretools.variable_types import Numeric, Text
 from nltk.tokenize.treebank import TreebankWordDetokenizer
 from sklearn.decomposition import TruncatedSVD
 from sklearn.feature_extraction.text import TfidfVectorizer
-from sklearn.pipeline import make_pipeline
+from sklearn.pipeline import make_pipeline, Pipeline
 
 from .utilities import clean_tokens
 
@@ -19,7 +19,9 @@ class LSA(TransformPrimitive):
         value decomposition to go from a sparse matrix to a compact matrix with two
         values for each string. These values represent that Latent Semantic Analysis
         of each string. These values will represent their context with respect to
-        the corpus of all strings in the given list.
+        (nltk's brown sentence corpus.)[https://www.nltk.org/book/ch02.html#brown-corpus].
+        Instead of the brown sentence corpus, this primitive may also be passed a
+        different LSA pipeline, pre-trained on a different corpus. 
 
         If a string is missing, return `NaN`.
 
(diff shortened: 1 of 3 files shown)


nlp_primitives/lsa.py
```diff
@@ -80,3 +87,25 @@ def lsa(array):
...
+    if not isinstance(trainer.steps[1][1], TruncatedSVD):
+        raise ValueError('Given LSA pipeline does not have a TruncatedSVD as its second transformer')
+    try:
+        trainer.steps[0][1].vocabulary_
+        trainer.steps[1][1].components_
+    except AttributeError:
+        raise ValueError('Given LSA pipeline should be pre-fitted')
+
+
+def make_trainer(corpus, random_state=42):
```
ctduffy, line 108: I think it would be really useful to see an example of using this in the docstring (either for this function, the LSA docstring, or both). Also, I think this function should probably be added to the `__init.py__` file so that users can access it through the package.

nlp_primitives/lsa.py
```diff
@@ -42,25 +44,41 @@ class LSA(TransformPrimitive):
...
+        resulting in a different output for the same input text.
+
+        >>> corpus = ['she will help eat food for a long time', 'I like to walk', 'We go for long walks together']
+        >>> trainer = make_trainer(corpus)
+        >>> lsa = LSA(trainer=trainer)
+        >>> x = ["he helped her walk,", "me me me eat food", "the sentence doth long"]
+        >>> res = lsa(x).tolist()
+        >>> for i in range(len(res)): res[i] = [abs(round(x, 2)) for x in res[i]]
+        >>> res
+        [[0.0, 0.34, 0.4], [0.58, 0.0, 0.0]]
```
dsherry, line 57: Awesome, thanks for including!

nlp_primitives/lsa.py
```diff
@@ -80,3 +99,34 @@ def lsa(array):
             return pd.Series(arr)
 
         return lsa
+
+
+def _validate_trainer(trainer):
```
dsherry, line 104: If we're expecting a specific pipeline structure here, instead of exposing a `trainer`parameter as input, I think we should expose a `training_corpus` parameter instead:
```python
lsa = LSA(training_corpus=corpus)
```

Then we can replace this validation with validation of the input data based on the expectations for that.

Which brings up a different question: what are the requirements on the input to TfidfVectorizer? I'm assuming it expects a single column containing text, can handle either pd dataframe or np array, etc.

Finally, I think we should define the validation/make methods inside the class, to show they're specific to this class

eccabay: I like the idea of passing in the corpus instead of the trainer object here. I didn't go with that option for this implementation due to featuretools' requirement that any values passed into the `__init__` of a primitive must be saved as a class attribute, which leads to unnecessary storage of the entire corpus.

I'm starting to lean towards moving this implementation to evalml instead. I don't like having these functions floating around outside the class, as there is no sort of structure in place for what they should look like yet. @rwedge would the same `__init__` --> class attribute requirement hold if this primitive was implemented outside of featuretools, or would there be any way to work around it? If so, that would be another bonus to moving this all to evalml instead.

dsherry: @eccabay ah, I didn't know about that requirement. Makes sense. In that case, I agree, let's move this into an LSA component in evalml, instead of making the changes here and using the primitive.

eccabay: Oops, I meant an instance attribute

eccabay: The new plan sounds good, I'll close this PR and move the implementation over to evalml.

PR #24, opened 2020-08-03
Create nlp_primitives[complete] install option and remove tensorflow from core requirements

thehomebrewnerd:
With this update several changes have been made:
- Remove `tensorflow` and tensorhub` as core requirements
- Create new `nlp_primitives[complete]` install option for using primitives that require tensorflow
- Create new test to verify error is raised when trying to use UniversalSentenceEncoder without having tensorflow installed
- Create new `installdeps` Makefile command for properly installing all development and testing requirements
- Update circleci `config.yml` to perform complete install for testing

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,4 +18,5 @@ unit-tests:
 .PHONY: installdeps
 installdeps:
 	pip install --upgrade pip
+	pip install -e .
 	pip install -r test-requirements.txt
--- a/Makefile
+++ b/Makefile
@@ -18,5 +18,4 @@ unit-tests:
 .PHONY: installdeps
 installdeps:
 	pip install --upgrade pip
-	pip install -e ".[complete]"
 	pip install -r test-requirements.txt
--- a/Makefile
+++ b/Makefile
@@ -14,3 +14,9 @@ lint-tests:
 
 unit-tests:
 	pytest --cache-clear --show-capture=stderr -vv
+
+.PHONY: installdeps
+installdeps:
+	pip install --upgrade pip
+	pip install -e ".[complete]"
+	pip install -r test-requirements.txt
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ nlp_primitives allows you to make use of text data in your machine learning pipe
 ### Install
 There are two options for installing nlp_primitives. Both of the options will also install Featuretools, if it is not already installed.
 
-The first option is to install a version of nlp_primitives that does not include Tensorflow. With this option, primitives that depend on Tensorflow cannot be used. To install nlp_primitives without Tensorflow run:
+The first option is to install a version of nlp_primitives that does not include Tensorflow. With this option, primitives that depend on Tensorflow cannot be used. Currently, the only primitive that can not be used with this install option is ``UniversalSentenceEncoder``. To install nlp_primitives without Tensorflow run:
 ```shell
 pip install nlp_primitives
 ```
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ nlp_primitives allows you to make use of text data in your machine learning pipe
 ### Install
 There are two options for installing nlp_primitives. Both of the options will also install Featuretools, if it is not already installed.
 
-The first option is to install a version of nlp_primtives that does not include Tensorflow. With this option, primitives that depend on Tensorflow cannot be used. To install nlp_primitives without Tensorflow run:
+The first option is to install a version of nlp_primitives that does not include Tensorflow. With this option, primitives that depend on Tensorflow cannot be used. To install nlp_primitives without Tensorflow run:
 ```shell
 pip install nlp_primitives
 ```
--- a/README.md
+++ b/README.md
@@ -7,8 +7,16 @@ nlp_primitives is a Python library with Natural Language Processing Primitives, 
 nlp_primitives allows you to make use of text data in your machine learning pipeline in the same pipeline as the rest of your data.
 
 ### Install
+There are two options for installing nlp_primitives. Both of the options will also install Featuretools, if it is not already installed.
+
+The first option is to install a version of nlp_primtives that does not include Tensorflow. With this option, primitives that depend on Tensorflow cannot be used. To install nlp_primitives without Tensorflow run:
 ```shell
-pip install 'featuretools[nlp_primitives]'
+pip install nlp_primitives
+```
+
+The second option is to install the complete version of nlp_primitives, which will also install Tensorflow and allow use of all primitives. To install the complete version of nlp_primitives:
+```shell
+pip install "nlp_primitives[complete]"
 ```
 
 ### Demos
--- a/nlp_primitives/tests/test_universal_sentence_encoder.py
+++ b/nlp_primitives/tests/test_universal_sentence_encoder.py
@@ -1,4 +1,9 @@
+import sys
+
 import pandas as pd
+import pytest
+
+from nlp_primitives import UniversalSentenceEncoder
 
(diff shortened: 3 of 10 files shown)


gsheni:
Should we put something in the readMe that specifies how to install the complete version and how to not install the complete version? Similar to [Featuretools](https://github.com/FeatureLabs/featuretools#add-ons)?

thehomebrewnerd:
> Should we put something in the readMe that specifies how to install the complete version and how to not install the complete version? Similar to [Featuretools](https://github.com/FeatureLabs/featuretools#add-ons)?

The way things are being set up currently, following the instructions in the readme and running `pip install featuretools[nlp_primitives]` will install the complete version of nlp primitives ([related featuretools pr](https://github.com/FeatureLabs/featuretools/pull/1103)). 

I was wondering if we want a featuretools extra that will not install the complete version? 

As it stands now, to install version of nlp primitives without tensorflow you would just run `pip install featuretools` and `pip install nlp_primitives` separately.

rwedge:
> As it stands now, to install version of nlp primitives without tensorflow you would just run `pip install featuretools` and `pip install nlp_primitives` separately.

`pip install nlp_primitives` would be sufficient

We should update the readme to mention how to install without tensorflow

thehomebrewnerd:
Just to add a little more detail. As it currently stands:
- Users can install nlp_primitives + featuretools without tensorflow by running `pip install nlp_primitives`
- Users can install nlp_primitives + featuretools + tensorflow by running either `pip install "nlp_primitives[complete]" or by running `pip install featuretools[nlp_primitives]`

rwedge:
Can we test serializing / deserializing this primitive to ensure it still works when loaded from disk?

thehomebrewnerd:
> Can we test serializing / deserializing this primitive to ensure it still works when loaded from disk?

Yes, I will look into this and add a test.

thehomebrewnerd:
> Can we test serializing / deserializing this primitive to ensure it still works when loaded from disk?

@rwedge I added a test that verifies serialization/deserialization of the primitive works in `test_universal_sentence_encoder.py`. Let me know if this is what you were thinking or if you wanted to see something else.

nlp_primitives/tests/test_universal_sentence_encoder.py
```diff
@@ -13,3 +18,16 @@ def test_regular(universal_sentence_encoder):
...
     assert a.equals(b)
+
+
+def test_without_tensorflow(universal_sentence_encoder):
+    # Simulate tensorflow being missing
+    tf_mod = sys.modules['tensorflow']
+    sys.modules['tensorflow'] = None
+    err_message = "In order to use the UniversalSentenceEncoder primitive install 'nlp_primitives[complete]'"
+    with pytest.raises(ImportError) as error:
+        UniversalSentenceEncoder()
```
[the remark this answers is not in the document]
thehomebrewnerd: @dsherry Actually, neither of these will generate the error. The error is only raised when initializing an instance of `UniversalSentenceEncoder()` as we aren't actually trying to import `tensorflow` or `tensorhub` until the `__init__` method on the primitive is called.

README.md
```diff
@@ -7,8 +7,16 @@ nlp_primitives is a Python library with Natural Language Processing Primitives,
 nlp_primitives allows you to make use of text data in your machine learning pipeline in the same pipeline as the rest of your data.
 
 ### Install
+There are two options for installing nlp_primitives. Both of the options will also install Featuretools, if it is not already installed.
+
+The first option is to install a version of nlp_primtives that does not include Tensorflow. With this option, primitives that depend on Tensorflow cannot be used. To install nlp_primitives without Tensorflow run:
```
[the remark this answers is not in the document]
thehomebrewnerd: Fixed

nlp_primitives/tests/test_universal_sentence_encoder.py
```diff
@@ -13,3 +18,16 @@ def test_regular(universal_sentence_encoder):
...
+    assert error.value.args[0] == err_message
+
+    # Add tensorflow back to sys.modules
+    sys.modules['tensorflow'] = tf_mod
```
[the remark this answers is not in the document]
thehomebrewnerd: Good catch. If something failed in the assertions, this would cause all downstream tests that need tensorflow to fail. I updated to move the change to `sys.modules` into a fixture as you suggested and that fixes the problem.

thehomebrewnerd:
> I think we should add the `[complete]` dependencies to `test-requirements.txt` since they are needed to run the tests

When I add the complete deps to `test-requirements.txt` I get an error when running `pip install`. This is because of the line below in `setup.py`:
```
tests_require=open('test-requirements.txt').readlines(),
```

If I remove this line, everything works as expected. I don't think this is necessary to include, but not sure why it was in there?

nlp_primitives/tests/test_universal_sentence_encoder.py
```diff
@@ -13,3 +23,63 @@ def test_regular(universal_sentence_encoder):
...
+    deserializer = PrimitivesDeserializer()
+    deserialized_primitive = deserializer.deserialize_primitive(serialized_primitive)
+
+    a = pd.DataFrame(deserialized_primitive(sentences))
+    a = a.mean().round(7).astype('str')
+    b = pd.Series(['-0.0007475', '0.0032088', '0.0018552', '0.0008256', '0.0028342'])
+    assert a.equals(b)
+
+
+def test_feature_serialization(universal_sentence_encoder):
```
rwedge, line 62: can this test use the tmpdir fixture to ensure file clean up

thehomebrewnerd: Updated

Makefile
```diff
@@ -14,3 +14,8 @@ lint-tests:
 
 unit-tests:
 	pytest --cache-clear --show-capture=stderr -vv
+
+.PHONY: installdeps
+installdeps:
+	pip install --upgrade pip
+	pip install -r test-requirements.txt
```
rwedge, line 21: i think this will fail to install the library, if that was the intention.  since the library is not mentioned in requirements.txt

thehomebrewnerd: Right - I think I'm getting myself confused with the different ways things can be installed.

Is this what we want for each of these?
- `make installdeps` - install everything, library, base requirements, test requirements, complete requirements
- `pip install -r requirements.txt` - install only base requirements, but not library (like featuretools)
- `pip install -r test-requirements.txt` - install only additional test and complete dependencies (similar to featuretools)
- `pip install -r complete-requirement.txt` - install only complete dependencies

rwedge: yeah that sums it up

thehomebrewnerd: Ok, I think the last commit should implement that behavior

README.md
```diff
@@ -7,8 +7,16 @@ nlp_primitives is a Python library with Natural Language Processing Primitives,
 nlp_primitives allows you to make use of text data in your machine learning pipeline in the same pipeline as the rest of your data.
 
 ### Install
+There are two options for installing nlp_primitives. Both of the options will also install Featuretools, if it is not already installed.
+
+The first option is to install a version of nlp_primitives that does not include Tensorflow. With this option, primitives that depend on Tensorflow cannot be used. To install nlp_primitives without Tensorflow run:
```
rwedge, line 12: I think for now, since only USE primitive depends on Tensorflow, we should mention it individually so the user knows its only one primitive they lose out on by skipping the Tensorflow install

thehomebrewnerd: Updated

PR #29, opened 2020-09-01
Refactor `clean_tokens` function

eccabay:
- Fixes a few `universal_sentence_encoder` tests that were failing on master
- Refactors the try/except/finally logic in `clean_tokens` so that a download failing will skip that part and complete successfully rather than raising a `Variable referenced before assignment` error
- Removes the `nltk.word_tokenizer` in `clean_tokens` altogether in preference for an equivalent set of operations that takes half the time to run
- Removes the `nltk` detokenizer from `PolarityScore`, since the same operation can be performed by `str.join(' ')` in a much shorter amount of time

Running cProfile:
Original `clean_tokens` performance
clean_tokens_original
After removing `word_tokenizer` 
clean_tokens_union_out_of_listcomp

### DIFF
--- a/nlp_primitives/polarity_score.py
+++ b/nlp_primitives/polarity_score.py
@@ -4,7 +4,6 @@ import pandas as pd
 from featuretools.primitives.base import TransformPrimitive
 from featuretools.variable_types import Numeric, Text
 from nltk.sentiment.vader import SentimentIntensityAnalyzer
-from nltk.tokenize.treebank import TreebankWordDetokenizer
 
 from .utilities import clean_tokens
 
@@ -29,7 +28,6 @@ class PolarityScore(TransformPrimitive):
     default_value = 0
 
     def get_function(self):
-        dtk = TreebankWordDetokenizer()
 
         def polarity_score(x):
             try:
@@ -50,6 +48,6 @@ class PolarityScore(TransformPrimitive):
                     if len(el) < 1:
                         li.append(0.0)
                     else:
-                        li.append(vader_pol(dtk.detokenize(el)))
+                        li.append(vader_pol(str.join(' ', el)))
             return pd.Series(li)
         return polarity_score
--- a/nlp_primitives/tests/test_universal_sentence_encoder.py
+++ b/nlp_primitives/tests/test_universal_sentence_encoder.py
@@ -1,6 +1,7 @@
 import sys
 
 import featuretools as ft
+import numpy as np
 import pandas as pd
 import pytest
 from featuretools.primitives.utils import (
@@ -20,9 +21,9 @@ def test_regular(universal_sentence_encoder):
         "Mitochondria is the powerhouse of the cell",
     ])
     a = pd.DataFrame(universal_sentence_encoder(sentences))
-    a = a.mean().round(7).astype('str')
-    b = pd.Series(['-0.0007475', '0.0032088', '0.0018552', '0.0008256', '0.0028342'])
-    assert a.equals(b)
+    a = a.mean().round(7).to_numpy()
+    b = pd.Series([-0.0007475, 0.0032088, 0.0018552, 0.0008256, 0.0028342])
+    np.testing.assert_array_almost_equal(a, b)
 
 
 @pytest.fixture()
@@ -54,9 +55,9 @@ def test_primitive_serialization(universal_sentence_encoder):
     deserialized_primitive = deserializer.deserialize_primitive(serialized_primitive)
 
     a = pd.DataFrame(deserialized_primitive(sentences))
-    a = a.mean().round(7).astype('str')
-    b = pd.Series(['-0.0007475', '0.0032088', '0.0018552', '0.0008256', '0.0028342'])
-    assert a.equals(b)
+    a = a.mean().round(7).to_numpy()
+    b = np.array([-0.0007475, 0.0032088, 0.0018552, 0.0008256, 0.0028342])
+    np.testing.assert_array_almost_equal(a, b)
 
 
 def test_feature_serialization(universal_sentence_encoder, tmpdir):
--- a/nlp_primitives/utilities.py
+++ b/nlp_primitives/utilities.py
@@ -8,24 +8,25 @@ wn = nltk.WordNetLemmatizer()
 
 def clean_tokens(textstr):
     try:
+        textstr = nltk.tokenize.word_tokenize(textstr)
+    except LookupError:
+        nltk.download('punkt')
+        textstr = nltk.tokenize.word_tokenize(textstr)
+
+    try:
         swords = set(nltk.corpus.stopwords.words('english'))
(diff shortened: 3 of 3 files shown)


nlp_primitives/polarity_score.py
```diff
@@ -50,6 +48,6 @@ def vader_pol(sentence):
                     if len(el) < 1:
                         li.append(0.0)
                     else:
-                        li.append(vader_pol(dtk.detokenize(el)))
+                        li.append(vader_pol(str.join(' ', el)))
```
ctduffy, line 51: good change here, this makes a lot of sense, bet it saved a good bit of time

nlp_primitives/tests/test_universal_sentence_encoder.py
```diff
@@ -20,9 +21,9 @@ def test_regular(universal_sentence_encoder):
         "Mitochondria is the powerhouse of the cell",
     ])
     a = pd.DataFrame(universal_sentence_encoder(sentences))
-    a = a.mean().round(7).astype('str')
-    b = pd.Series(['-0.0007475', '0.0032088', '0.0018552', '0.0008256', '0.0028342'])
-    assert a.equals(b)
+    a = a.mean().round(7).to_numpy()
+    b = pd.Series([-0.0007475, 0.0032088, 0.0018552, 0.0008256, 0.0028342])
```
ctduffy, line 25: good catch here, weird that I (at least I think I was the culprit here) had cast it to a string to compare.

nlp_primitives/utilities.py
```diff
@@ -7,25 +6,25 @@
 
 
 def clean_tokens(textstr):
+    textstr = textstr.translate(str.maketrans('', '', string.punctuation))
```
ctduffy, line 9: is there a reason to use this instead of just replacing all punctuation with an empty string with regex maybe? This could be a better way/not the only reason for this line (it seems like this line is meant to remove all punctuation, but feel free to correct me), but just a thought/wondering the rationale for this construction in particular. I see that you have removed the `import re` line so a reason could be that regex was really slow?

nlp_primitives/utilities.py
```diff
@@ -7,25 +6,25 @@
...
+    textstr = [ch for ch in textstr.split(' ') if len(ch) > 0]
+
     try:
         swords = set(nltk.corpus.stopwords.words('english'))
+        to_remove = set(string.punctuation).union(swords)
+        textstr = [ch.lower() for ch in textstr if ch not in to_remove]
     except LookupError:
         nltk.download('stopwords')
         swords = set(nltk.corpus.stopwords.words('english'))
+        to_remove = set(string.punctuation).union(swords)
```
ctduffy, line 19: might it be worth not repeating this and the following line twice as they are the same as on 14 and 15, and could instead be taken out of the `try`/`except` and just put after the `except` expression and not be indented? / could be put in a `finally` expression

nlp_primitives/utilities.py
```diff
@@ -7,25 +6,25 @@
...
-                wn.lemmatize("")
-            except LookupError:
-                nltk.download('wordnet')
-            finally:
-                processed = [ch.lower() for ch in textstr if ch not in
-                             set(string.punctuation).union(swords)]
-                processed = ['0' if re.search('[0-9]+', ch) else ch for ch in processed]
-                processed = [wn.lemmatize(w) for w in processed]
-                return processed
+        textstr = ['0' if any(map(str.isdigit, ch)) else ch for ch in textstr]
```
ctduffy, line 29: similar comment to above (in removing punctuation), just wondering what the reason is for moving away from regex, but I do think this looks very clean!

eccabay: I removed the regex here to add just a little more optimization - the `re.search` runtime is the third largest box in the second cProfile screenshot I posted, it took about 50 seconds. Removing the regex sped things up noticeably, the same operation was cut down to 20 seconds, so `clean_tokens` is down to running in 340 seconds total.

Apparently, the way regex searches through text includes a lot of backtracking and jumping forward, and it can be really inefficient on some types of datasets, depending on what you're searching for (https://towardsdatascience.com/regex-performance-in-python-873bd948a0ea), so I figured it would be safer to pull them out of here at least for now.

nlp_primitives/utilities.py
eccabay: Everything I've found online indicates that this `maketrans` is just about the most efficient way to do this operation, apparently it just directly wraps C code, which is nice. See my comment below about regex and why I moved away from that.

nlp_primitives/utilities.py
eccabay: There's a [weird bug ]( https://github.com/FeatureLabs/evalml/issues/1083)going on in the `evalml` code right now that actually stems from this. If the internet connection is bad or non-existent when running this code, the nltk download will fail and any attempts to use that variable will throw an error. By keeping all references to nltk's stopwords within the try/except block, the code will run and successfully return something even if the download(s) fail.

nlp_primitives/utilities.py
ctduffy: amazing, makes a lot of sense, cool that you looked into this!

nlp_primitives/utilities.py
ctduffy: Hm that is strange, I guess there isn't a lot of python handling around downloading and waiting for stuff to download/handling if it fails in the evalml library/here. in that case, this seems very reasonable!!

nlp_primitives/utilities.py
ctduffy: oooh cool!! the more ya know!

nlp_primitives/tests/test_universal_sentence_encoder.py
```diff
@@ -54,9 +55,9 @@ def test_primitive_serialization(universal_sentence_encoder):
     deserialized_primitive = deserializer.deserialize_primitive(serialized_primitive)
 
     a = pd.DataFrame(deserialized_primitive(sentences))
-    a = a.mean().round(7).astype('str')
-    b = pd.Series(['-0.0007475', '0.0032088', '0.0018552', '0.0008256', '0.0028342'])
-    assert a.equals(b)
+    a = a.mean().round(7).to_numpy()
+    b = np.array([-0.0007475, 0.0032088, 0.0018552, 0.0008256, 0.0028342])
+    np.testing.assert_array_almost_equal(a, b)
```
dsherry, line 60: Looks good.

dsherry (approved):
👍  LGTM! I know we've got other changes planned which will touch the download-related code, so I'll hold a few comments until we get to that work.

dsherry:
@eccabay any reason not to merge this?

eccabay:
@dsherry nope, not really! Was holding on in case anyone had anything else to add, but I'll just go ahead and merge now

PR #32, opened 2020-09-24
Speedups to polarity score and other primitives

dsherry:


dsherry:
Covered by #31 , closing

PR #34, opened 2020-10-08
Update to 0.20.0 Featuretools and Support new NaturalLanguage variable type 

### LINKED ISSUE #33: update nlp primitives to support new NaturalLanguage variable type in featuretools
Featuretools 0.20.0 deprecated the Text variable type in favor of NaturalLanguage variable type.  The current NLP primitives use Text variables as input, and as such will not be applied to Natural Language variables.  We should update the primitives to use Natural Language input, which works with both Text and Natural Language input.

gsheni:
- Fixes #33

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -3,8 +3,16 @@ Changelog
 =========
 **v1.0.0** Aug 12, 2020
     * Changes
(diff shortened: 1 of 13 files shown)


rwedge (approved):
Looks good

PR #36, opened 2020-10-19
Move nltk data tarball to git-lfs

dsherry:
Links:
* https://docs.github.com/en/free-pro-team@latest/github/managing-large-files/installing-git-large-file-storage
* https://docs.github.com/en/free-pro-team@latest/github/managing-large-files/configuring-git-large-file-storage


dsherry:
No longer necessary after #39

PR #39, opened 2020-10-22
Include nltk data uncompressed; restore support for pip-installing specific commits

dsherry:
**Problem**
#31 made it so that during sdist creation, nltk data files are unpacked from a tarfile and included in the `nltk` data path.

This works great, but introduces the following limitation: when `nlp_primitives` is installed in a way *other* than from an sdist/wheel/egg, the nltk data files aren't unpacked. This is the case for the following install paths:
* Install from repo source: `pip install .`
* Install from specific hash: `pip install -e git+git@github.com:FeatureLabs/nlp_primitives.git@<hash>#egg=nlp_primitives`, where `<hash>` is a git commit hash

The former case was handled by the additions we made to the makefile to unpack the tarball during developer env setup. But the latter case cannot be addressed in this manner. We've encountered this as we prepare to use the nlp primitives in evalml--we want to be able to pip-install a specific hash from this repo in our testing, and find we're unable to do so.

**Fix**
I looked into hijacking other setuptools commands [like `install`](https://github.com/pypa/setuptools/blob/master/setuptools/command/install.py) and `develop`. This would work, but I wasn't able to get the pathing to work out correctly. When I ran `pip install .`, I saw the nltk data tarball was getting extracted to a temp directory instead of to `site-packages` in the virtualenv I was in.

Instead of digging further into that approach, I suggest we simply add the data here in uncompressed format. The compression wasn't saving any space -- the tarball and the unpacked directory are both currently around 53M.

**Future**
Rather than always including the data files in the package, we may want to look into providing both an "offline" installation path, where the data is included in the package, and an "online" installation path, where the data is downloaded during installation.


rwedge (commented):
we should probably update the line in the readme about unpacking the nltkd ata

dsherry:
@rwedge good call, done!

rwedge (approved):
looks good

PR #42, opened 2020-10-26
Update universal_sentence_encoder doctest

rwedge:
Updates the docstring example to use the `universal_sentence_encoder` fixture that other tests do.  Appears to fix problems with tests crashing during the Universal Sentence Encoder tests


frances-h (approved):
looks good!

PR #43, opened 2020-10-26
v1.1.0

rwedge:
**v1.1.0** Oct 26, 2020
* Changes
    * Upgrade to Featuretools 0.20.0 and replace Text variable type with NaturalLanguage (#34)
    * Include corpus download with package instead of downloading at first run. (#31)
    * Change corpus used by LSA primitive (#35)
    * Refactor ``clean_tokens`` function (#29)
* Testing Changes
    * Add another set of CI jobs which ensure that we can make an nlp_primitives package, install that, and then still pass all the unit tests, vs an editable install (#31)
    * Move some test utils into `test/` (#31)


dsherry (approved):
Great! 🚢

PR #44, opened 2020-11-02
Use authenticated docker pulls for Circle CI

rwedge:
Update Circle CI config to use the job matrix format and authenticate with dockerhub


gsheni (approved):
lgtm

PR #45, opened 2020-11-04
Replace master references with main

rwedge:
Still some residual references due to sphinx

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -8,9 +8,9 @@ jobs:
     name: PyPI Release
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@master
+    - uses: actions/checkout@v2
     - name: PyPI Upload
-      uses: FeatureLabs/gh-action-pypi-upload@master
+      uses: FeatureLabs/gh-action-pypi-upload@v1
       env:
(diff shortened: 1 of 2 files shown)


ctduffy (changes requested):
I see a couple more appearances of `master` in the release.md file. They are in the "Set up fork of our conda-forge repo" section. Should this be changed, or is the branch still called master in the conda-forge repo? (If this is the case, should we change this to main?)

rwedge:
> I see a couple more appearances of `master` in the release.md file. They are in the "Set up fork of our conda-forge repo" section. Should this be changed, or is the branch still called master in the conda-forge repo? (If this is the case, should we change this to main?)

Yes, the branch is still called master in the conda-forge repo.   There would need to be some research on how closely tied conda-forge is to the name of the default branch before changing it

ctduffy (approved):
> > I see a couple more appearances of `master` in the release.md file. They are in the "Set up fork of our conda-forge repo" section. Should this be changed, or is the branch still called master in the conda-forge repo? (If this is the case, should we change this to main?)
> 
> Yes, the branch is still called master in the conda-forge repo. There would need to be some research on how closely tied conda-forge is to the name of the default branch before changing it

Ok, sounds good, maybe that is something we can look into later, I am not too familiar with conda-forge, so I am not sure how much work it takes to change to main there

PR #46, opened 2020-12-23
Migrate Tests

jeff-hernandez:
Tests were migrated to GitHub Actions.


rwedge (commented):
The scheduler is new test content - perhaps we should have a PR that only ports the existing tests first.

jeff-hernandez:
Sure, I can separate the scheduler into another PR

rwedge:
When we do add the scheudre, I think we'll need to add the featuretools release/main concept in order  to get the most usefulness out of it

jeff-hernandez:
Sounds good. We may need to update the circleci action.

rwedge (approved):
Looks good

PR #48, opened 2021-01-20
Run tests on recent commits to Featuretools

jeff-hernandez:
The check for recent commits is scheduled to happen every day at 12 pm EST


rwedge (commented):
How will recent commits to Featuretools impact the tests in this repo?

jeff-hernandez:
Recent commits to Featuretools that occur within the past 25 hours will trigger the test workflow in this repo to run. Also, we should see that the test workflow was manually run by machineFl in the Actions tab.

Screen Shot 2021-01-21 at 11 26 09 AM

rwedge:
I don't think the current tests in this repo install featuretools from the github repo, so I don't think they'd be impacted by updates to featuretools main branch

We'd need to update the tests to like how we do it for the tsfresh primitives if we wanted that kind of testing

jeff-hernandez:
Ah, that makes sense. I'll update the test workflow to be like the one for tsfresh primitives.

rwedge (approved):
LGTM

rwedge:
Added the token secret to the repo

PR #49, opened 2021-01-21
Test using latest version of Featuretools 

jeff-hernandez:
The test workflow creates unit tests that install the latest version of Featuretools from the main branch. Branch protection rules should include the additional tests.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -28,7 +28,7 @@ jobs:
           make installdeps-complete
       - if: ${{ matrix.featuretools_version == 'Main' }}
         name: Install latest version of Featuretools from main branch
-        run: pip install git+https://github.com/FeatureLabs/featuretools.git@main#egg=featuretools --upgrade
+        run: pip install git+https://github.com/alteryx/featuretools.git@main#egg=featuretools --upgrade
       - name: Test entry point
         run: make entry-point-test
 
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -25,6 +26,9 @@ jobs:
         run: |
           pip config --site set global.progress_bar off
           make installdeps-complete
+      - if: ${{ matrix.featuretools_version == 'Main' }}
+        name: Install latest version of Featuretools from main branch
+        run: pip install git+https://github.com/alteryx/featuretools.git@main#egg=featuretools --upgrade
```
rwedge, line 31: I think pip is skipping installing `main` if the version numbers match. 

```
make installdeps-test
pip install git+https://github.com/alteryx/featuretools.git@main#egg=featuretools --upgrade
python -c "from featuretools.tests.entityset_tests.test_dask_es import test_single_table_dask_entityset"
```
runs cleanly while

```
pip install --upgrade pip
pip install git+https://github.com/alteryx/featuretools.git@main#egg=featuretools --upgrade
pip install pytest
python -c "from featuretools.tests.entityset_tests.test_dask_es import test_single_table_dask_entityset"
```
gives 

```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
ImportError: cannot import name 'test_single_table_dask_entityset' from 'featuretools.tests.entityset_tests.test_dask_es'
```
Which makes sense, because that test was moved out of that file since the last release of featuretools

jeff-hernandez: Good catch I'll use the force reinstall option

rwedge (approved):
looks good!

PR #61, opened 2021-08-11
[DO NOT MERGE] Implements changes for compatibility with Featuretools 1.0.0

thehomebrewnerd:
Implements changes for compatibility with Featuretools 1.0.0

This PR should not be merged until Featuretools 1.0.0 is being released.

### DIFF
--- a/nlp_primitives/mean_characters_per_word.py
+++ b/nlp_primitives/mean_characters_per_word.py
@@ -3,7 +3,8 @@
 import re
 
 from featuretools.primitives.base import TransformPrimitive
-from featuretools.variable_types import NaturalLanguage, Numeric
+from woodwork.column_schema import ColumnSchema
+from woodwork.logical_types import NaturalLanguage
 
 
 class MeanCharactersPerWord(TransformPrimitive):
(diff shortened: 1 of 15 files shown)


thehomebrewnerd:
Note - the tests with Featuretools `main` and the tests with Featuretools latest release will continue to fail until Featuretools 1.0.0 is released. In order for the tests to pass in the meantime, Featuretools needs to be installed from the `woodwork-integration` branch.

nlp_primitives/mean_characters_per_word.py
```diff
@@ -39,6 +40,6 @@ def mean_characters_per_word(x):
             df = df.reset_index()
             df = df.melt(id_vars='index', var_name='word_index', value_name='word')
             df['n_characters'] = df['word'].str.len()
-            results = df.groupby(by=['index']).mean()['n_characters']
+            results = df.groupby(by=['index']).mean(numeric_only=False)['n_characters']
```
thehomebrewnerd, line 43: This change is required because in situations where the column of words contains `pd.NA` values the `n_characters` column has a dtype of `object` which will not have `mean` calculated by default.

davesque (dismissed):
Looks good to me!  I assume tests are passing locally over the featuretools dev branch?

thehomebrewnerd:
> Looks good to me! I assume tests are passing locally over the featuretools dev branch?

Yes, correct.

tamargrey:
From @rwedge: Specifying the logical type for primitives we know will always return integers or floats would help avoid inference errors / reduce infer time (as opposed to only numeric tags for all numeric primitives)

(seems relevant here as well since a lot of the primitives are floating point scores or integer counts)

rwedge (approved):

PR #79, opened 2021-12-15
Add primitive for total word length

jeff-hernandez:


nlp_primitives/total_word_length.py
```diff
@@ -0,0 +1,48 @@
...
+        >>> total_word_length = TotalWordLength()
+        >>> total_word_length(x).tolist()
+        [15.0, 16.0, 15.0]
+    """
+    name = "total_word_length"
+    input_types = [ColumnSchema(logical_type=NaturalLanguage)]
+    return_type = ColumnSchema(logical_type=Double, semantic_tags={'numeric'})
+    default_value = 0
+
+    def __init__(self, delimiters_regex=None):
```
rwedge, line 41: do we want to use a different default regex?

thehomebrewnerd:
If I'm understanding this correctly, all we are doing it counting the number of characters in a string that are not delimiter characters. WIth that in mind, could we simplify this significantly and make the primitive function something like this instead and not do all the tokenization stuff?

```
def get_function(self):
    def total_word_length(x):
        counts = x.str.len() - x.str.findall(r"[- \[\].,!\?;\n]").apply(lambda x: len(x))
        return counts
    return total_word_length
```

The pattern `r"[- \[\].,!\?;\n]"` in the `findall` call can specify all the delimiters we don't want to count.

nlp_primitives/total_word_length.py
gsheni: Yes, we should use a default delimiter regex if we can, rather than `None`

gsheni:
@thehomebrewnerd if that implementation is faster, let's go ahead and do that.

nlp_primitives/total_word_length.py
jeff-hernandez: Are we okay with using `[- \[\].,!\?;\n]` for default delimiters?

gsheni: yes, let's make sure we explain this in the docstring (perhaps by listing them out, bullets?)

jeff-hernandez: Another set of characters that `MeanCharactersPerWord` uses is `!,.:;?`

nlp_primitives/total_word_length.py
```diff
@@ -0,0 +1,41 @@
...
+    return_type = ColumnSchema(logical_type=Double, semantic_tags={'numeric'})
+
+    default_value = 0
+
+    def __init__(self, delimiters_regex=r"[- \[\].,!\?;\n]"):
+        self.delimiters_regex = delimiters_regex
+
+    def get_function(self):
+        def total_word_length(x):
+            x = x.str.len() - x.fillna(str()).str.findall(self.delimiters_regex).apply(len)
```
jeff-hernandez, line 39: Using `fillna(str())` avoids raising an error when calling `str.findall` on null values

nlp_primitives/total_word_length.py
```diff
@@ -0,0 +1,41 @@
...
+
+    Examples:
+        >>> x = ['This is a test file', 'This is second line', 'third line $1,000', None]
+        >>> total_word_length = TotalWordLength()
+        >>> total_word_length(x).tolist()
+        [15.0, 16.0, 14.0, nan]
+    """
+    name = "total_word_length"
+    input_types = [ColumnSchema(logical_type=NaturalLanguage)]
+    return_type = ColumnSchema(logical_type=Double, semantic_tags={'numeric'})
```
jeff-hernandez, line 30: In this implementation, if the series does not contain null values, it'll return an integer dtype otherwise float. If we want to return floats, then we'll need to convert the series dtype in those cases.

nlp_primitives/total_word_length.py
```diff
@@ -36,6 +36,7 @@ def __init__(self, delimiters_regex=r"[- \[\].,!\?;\n]"):
 
     def get_function(self):
         def total_word_length(x):
-            x = x.str.len() - x.fillna(str()).str.findall(self.delimiters_regex).apply(len)
-            return x
+            delimiters = x.fillna(str()).str.findall(self.delimiters_regex)
+            counts = x.str.len() - delimiters.apply(len)
```
tamargrey, line 40: Would you be able to do `x.str.count(self.delimiters_regex)` here?

I think it's slightly faster than having to findall and then check the length
image

jeff-hernandez: yep, that works -- updated

tamargrey (dismissed):
This looks good to me. Not sure what sort of testing should be happening--if it would make sense to add something like https://github.com/alteryx/nlp_primitives/blob/main/nlp_primitives/tests/test_punctuation_count.py

But if this primitive is covered by other tests, happy to keep as is.

nlp_primitives/total_word_length.py
thehomebrewnerd: @tamargrey Good call!

nlp_primitives/total_word_length.py
thehomebrewnerd: I don't think this should matter much. By specifying the return type as `Double` here, things will get converted to a float when Woodwork is initialized on the feature matrix.

thehomebrewnerd (commented):
I think we might want to add a unit test or two, beyond the docstring test, just to make sure things work as expected in different scenarios. A couple that come to mind:
- user overrides the default regex string
- A string containing only delimiters
- A series containing different representations for missing values: np.nan, None, pd.NA

nlp_primitives/total_word_length.py
jeff-hernandez: If type conversions happen only after applying all the primitives, aggregation primitives can get unexpected inputs types. For example, an aggregation primitive after this transform primitive can get an integer type in some cases even when the input type is double.

thehomebrewnerd: Yeah, technically I think that could be an issue, but as of now I don't think we have any primitives that specify a specific numeric logical type (Double, Integer, IntegerNullable, etc) as input, do we? I think everything that takes in a numeric value specifies the input like this: 
```
input_types = [ColumnSchema(semantic_tags={'numeric'})]
```

If we don't know the logical type exactly, you can specify the return type without the logical type, which is probably more consistent with the actual behavior in this case:
```
return_type = ColumnSchema(semantic_tags="numeric")
```

jeff-hernandez: Yeah, I'm mainly just highlighting the unexpected input types. It doesn't have to be numeric. If a transform primitive returns an unknown type when the return type is double, then that could raise issues.

nlp_primitives/total_word_length.py
```diff
@@ -0,0 +1,41 @@
...
+        >>> x = ['This is a test file', 'This is second line', 'third line $1,000', None]
+        >>> total_word_length = TotalWordLength()
+        >>> total_word_length(x).tolist()
+        [15.0, 16.0, 14.0, nan]
+    """
+    name = "total_word_length"
+    input_types = [ColumnSchema(logical_type=NaturalLanguage)]
+    return_type = ColumnSchema(logical_type=Double, semantic_tags={'numeric'})
+
+    default_value = 0
```
gsheni, line 32: Is this needed?

thehomebrewnerd (approved):
LGTM!

thehomebrewnerd (approved):

PR #80, opened 2021-12-17
Add primitive for median word length

jeff-hernandez:


nlp_primitives/median_word_length.py
```diff
@@ -0,0 +1,46 @@
...
+        self.delimiters_regex = delimiters_regex
+
+    def get_function(self):
+        def get_median(words):
+            if isinstance(words, list):
+                return median([len(word) for word in words])
+
+        def median_word_length(x):
+            words = x.str.split(self.delimiters_regex)
+            return words.apply(get_median)
```
thehomebrewnerd, line 45: Not sure if it would be possible or more efficient to do this in a single `apply` function call? Just wondering a bit if we are traversing the series values twice here - once to split the strings into a list of words and then a second time to get the median length value for each list. Could we do this in a single function that does both the split and the median calculation?

I'm not entirely sure which would perform better, so we might need to test out both approaches on a long series to see what is faster.

nlp_primitives/tests/test_median_word_length.py
```diff
@@ -0,0 +1,42 @@
...
+
+        expected = pd.Series([4., 4.5, 8.])
+        actual = self.primitive('[ ,;]').get_function()(x)
+        pd.testing.assert_series_equal(actual, expected, check_names=False)
+
+    def test_multiline(self):
+        x = pd.Series(['This is a test file.',
+                       'This is second line\nthird line $1000;\nand subsequent lines'])
+
+        expected = pd.Series([3.0, 4.0])
```
thehomebrewnerd, line 24: I think we are getting some empty strings during the splitting when we have consecutive delimiters that are impacting the results. I think the median word length for `This is a test file.` should be 4, not 3, no? There are 5 words with lengths `1, 2, 4, 4, 4`.

Maybe need to update `[len(word) for word in words]` to `[len(word) for word in words if len(word) != 0]` in the `get_median` function.

thehomebrewnerd (commented):
Also, I forgot to mention this on the last PR, but we should add this update to the changelog in the docs folder along with the primitive you added last week.

nlp_primitives/median_word_length.py
jeff-hernandez: After comparing both approaches, it seems the first is a bit faster. The underlying implementation in pandas is probably better optimized for traversing the series.

Screen Shot 2021-12-20 at 10 04 15 AM

nlp_primitives/tests/test_median_word_length.py
jeff-hernandez: yeah, makes sense --updated

nlp_primitives/median_word_length.py
thehomebrewnerd: Ok, cool. Thanks for looking into it!

thehomebrewnerd (approved):
LGTM!

PR #84, opened 2021-12-21
Remove conda forge testing in release documentation

gsheni:


docs/source/changelog.rst
```diff
@@ -1,7 +1,18 @@
 =========
 Changelog
 =========
+.. Future Release
```
thehomebrewnerd, line 4: Uncomment this. Also if we are going to use the underline approach for the release version/date like we do in other repos, should we update the past entries so the formatting is consistent? Not a big deal either way though.

rwedge, line 4: this should be uncommented since the release already happened

thehomebrewnerd (dismissed):
One minor thing, but otherwise looks fine.

docs/source/changelog.rst
gsheni: Done

gsheni: Eh, let's leave it for now.
https://github.com/alteryx/nlp_primitives/issues/85

gsheni: Actually done.
https://github.com/alteryx/nlp_primitives/pull/84/commits/1deb85a86fce48b0a922260657d04c726968ab14

rwedge (approved):

thehomebrewnerd (approved):

PR #92, opened 2022-01-04
Add NumberOfCommonWords primitive

### LINKED ISSUE #74: Add NumberOfCommonWords primitive
- Given a list of 100/1000 most common words, determine how many of those exist.

```python
class NumberOfCommonWords(TransformPrimitive):
    name = "number_of_common_words"
    input_types = [ColumnSchema(logical_type=NaturalLanguage)]
    return_type = ColumnSchema(logical_type=Integer, semantic_tags={'numeric'})
```

## How to get common words
- https://en.wikipedia.org/wiki/Most_common_words_in_English
- https://gist.github.com/deekayen/4148741

tuethan1999:
Fixes #74


nlp_primitives/constants.py
```diff
@@ -0,0 +1,12 @@
...
+
+wikipedia_top_100 = pd.DataFrame({
```
thehomebrewnerd, line 5: Why is `wikipedia_top_100` included here? Doesn't seem to be used anywhere else or am I missing something?

nlp_primitives/number_of_common_words.py
```diff
@@ -0,0 +1,53 @@
...
+from woodwork.column_schema import ColumnSchema
+from woodwork.logical_types import IntegerNullable, NaturalLanguage
+
+from .constants import common_words_1000
+
+
+class NumberOfCommonWords(TransformPrimitive):
+    """Determines the number of common words in a string.
+    Description:
+        Given list of strings, determine the number of words that appear in a supplied wordbank.
```
thehomebrewnerd, line 14: Probably would be good to clarify that the word bank is optional and mention what gets used by default if one is not supplied.

nlp_primitives/number_of_common_words.py
```diff
@@ -0,0 +1,53 @@
...
+        >>> number_of_common_words(x).tolist()
+        [2, 1, 3]
+    """
+    name = "number_of_common_words"
+    input_types = [ColumnSchema(logical_type=NaturalLanguage)]
+    return_type = ColumnSchema(logical_type=IntegerNullable, semantic_tags={'numeric'})
+
+    default_value = 0
+
+    def __init__(self, word_set=set(common_words_1000), delimiters_regex=r"[- \[\].,!\?;\n]"):
```
thehomebrewnerd, line 30: I see we have the option to specify `delimiters_regex` here. We should mention it in the docstring as well, like we do the word bank, explain how it is used and add a doctest example that shows how to use it.

nlp_primitives/number_of_common_words.py
```diff
@@ -0,0 +1,53 @@
...
+                    num_common_words += 1
+            return num_common_words
+
+        def num_common_words(x):
+            assert (
+                self.delimiters_regex is not None
+            ), "delimiters_regex needs to be defined"
+            assert (
+                self.word_set is not None
+            ), "word_set needs to be defined"
```
thehomebrewnerd, line 50: Since we have default values for both, I'm not sure these assertions are necessary?

nlp_primitives/tests/test_number_of_common_words.py
```diff
@@ -0,0 +1,50 @@
+import numpy as np
+import pandas as pd
+
+from ..number_of_common_words import NumberOfCommonWords
+from .test_utils import PrimitiveT, find_applicable_primitives, valid_dfs
+
+
+class TestNumberOfCommonWords(PrimitiveT):
+    primitive = NumberOfCommonWords
+    test_word_bank = {'the', 'of', 'to', 'and', 'a', 'in', 'is', 'it', 'you', 'that'}
```
thehomebrewnerd, line 10: Can we make this simpler and only include the values that we are matching on in the tests? That would make it a little easier to determine the expected results manually.

thehomebrewnerd (commented):
No major issues here, but I have a few suggestions. We should also add this update to the changelog file.

nlp_primitives/number_of_common_words.py
tamargrey: Could add this and `word_set` to an `Args` section  of the docstring

nlp_primitives/constants.py
tuethan1999: It's not used for anything but it was mentioned in #74.  @gsheni thoughts?

gsheni: We can remove it. I don't remember specifically why I mentioned it in the issue.

nlp_primitives/number_of_common_words.py
tuethan1999: I was thinking of the case where the user sets the value to None. I've deleted them

nlp_primitives/number_of_common_words.py
```diff
@@ -0,0 +1,59 @@
...
+
+from .constants import common_words_1000
+
+
+class NumberOfCommonWords(TransformPrimitive):
+    """Determines the number of common words in a string.
+    Description:
+        Given list of strings, determine the number of words that appear in a supplied wordbank.
+        The wordbank defaults to nlp_primitives.constants.common_words_1000. The list of strings
+        is case insensitive. The word banks should be all lower case strings. If a string is
```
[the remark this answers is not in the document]
tuethan1999: I updated the docstring to reflect these changes!

tamargrey (approved):
lgtm

Issue #73, opened 2021-12-03
Add TokenCount primitive


gsheni:
- This primitive converts a collection of text documents to a matrix of token counts.
- We can use sklearn's feature extractor
  - https://scikit-learn.org/stable/modules/generated/sklearn.feature_extraction.text.CountVectorizer.html#sklearn.feature_extraction.text.CountVectorizer

PR #94, opened 2022-01-06
Add TokenCount primitive

tuethan1999:
#73


tuethan1999:
There's the potential for this primitive to create a LOT of output features since potentially every word can get tokenized

thehomebrewnerd:
> There's the potential for this primitive to create a LOT of output features since potentially every word can get tokenized

I am a little concerned by this behavior. Is this what we really want? I tested this primitive on a dataset with restaurant reviews and this primitive created a single feature (`<Feature: TOKEN_COUNT(Review)>`) that added 1,969 columns to  the feature matrix.

image

gsheni:
Let's discuss this primitive at standup.

tuethan1999:
@alteryx/ml-tools decided that the large number of features this primitive produces isn't worth the functionality

PR #95, opened 2022-01-07
Update scheduler workflow to include token

gsheni:


.github/workflows/scheduler.yml
```diff
@@ -17,6 +17,7 @@ jobs:
           repository: alteryx/nlp_primitives
           branch: main
           workflow: Unit Tests - Latest Dependencies
+          token: ${{ secrets.REPO_SCOPED_TOKEN }}
```
jeff-hernandez, line 20: Do we really need this token? `featuretools-tsfresh-primitives` successfully ran this kind of workflow recently without using a token ([workflow run](https://github.com/alteryx/featuretools-tsfresh-primitives/actions/runs/1668523694)).

gsheni: There was a failure:
https://github.com/alteryx/nlp_primitives/actions/runs/1668523390

jeff-hernandez: There doesn't seem to be a workflow run on main for `Unit Tests - Latest Dependencies` in the actions tab. I manually ran that workflow just now. Let's wait until it completes successfully before running this one again. We won't need the toke for this specific step, only for the last step to trigger the other workflow.

gsheni: There is though:
https://github.com/alteryx/nlp_primitives/blob/main/.github/workflows/unit_tests_with_latest_deps.yml#L9

jeff-hernandez: I mean the log of the workflow in [this actions tab](https://github.com/alteryx/nlp_primitives/actions/workflows/unit_tests_with_latest_deps.yml?query=branch%3Amain). It's appearing now though, so my browser probably wasn't finished loading.

Hmm, it seems it working now based on [this run](https://github.com/alteryx/nlp_primitives/actions/runs/1668683381). I'm not entirely sure what the problem was.

PR #98, opened 2022-02-09
Add entry point for featuretools primitives

jeff-hernandez:
After we merge [this PR in featuretools](https://github.com/alteryx/featuretools/pull/1891) and release the next version of Featuretools, the entry point tests should pass.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,7 @@
 entry-point-test:
 	python -c "from featuretools import nlp_primitives"
 	python -c "from featuretools.primitives import Elmo"
+	python -c "from nlp_primitives import Elmo"
 
 .PHONY: clean
 clean:
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,7 @@
 .PHONY : entry-point-test
 entry-point-test:
 	python -c "from featuretools import nlp_primitives"
+	python -c "from featuretools.primitives import Elmo"
 
 .PHONY: clean
 clean:
--- a/nlp_primitives/tests/test_utils.py
+++ b/nlp_primitives/tests/test_utils.py
@@ -10,7 +10,7 @@ import featuretools as ft
 import pytest
 from featuretools import dfs, list_primitives
 from featuretools.tests.testing_utils import make_ecommerce_entityset
-
+ft.primitives._load_primitives()
 
 class PrimitiveT:
     primitive = None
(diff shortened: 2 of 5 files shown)


Makefile
```diff
@@ -1,6 +1,7 @@
 .PHONY : entry-point-test
 entry-point-test:
 	python -c "from featuretools import nlp_primitives"
+	python -c "from featuretools.primitives import Elmo"
```
gsheni, line 4:
```suggestion
	python -c "from featuretools.primitives import Elmo"
	python -c "from nlp_primitives import Elmo"
```

gsheni: Put this in because EvalML calls it like this:
https://github.com/alteryx/evalml/blob/bb0f4f7f6a874916f73d2c8982f014828fe91dd7/evalml/pipelines/components/transformers/preprocessing/natural_language_featurizer.py#L6

rwedge:
I'm a bit confused as to why this code block still works, since the entry point should be adding these primitives to `list_primitives`:

https://github.com/alteryx/nlp_primitives/blob/a2066e71642a89108f587bdfa351050aa8f9f2e6/nlp_primitives/tests/test_utils.py#L34-L38

From what I can tell, during the tests `list_primitives` does not include the nlp primitives

@jeff-hernandez can you look into this further?

jeff-hernandez:
@rwedge importing `nlp_primitives` first will leave the custom primitives out of scope because the featuretools entry point runs too early -- loading `nlp_primitives` before any of the custom primitives have been defined.

```python
In [1]: import nlp_primitives
   ...: import featuretools as ft
   ...: len(ft.list_primitives())
   ...: 
Out[1]: 95
```

Importing featuretools first will run the entry point and completely load all the custom primitives into scope.

```python
In [1]: import featuretools as ft
   ...: import nlp_primitives
   ...: len(ft.list_primitives())
   ...: 
Out[1]: 106
```

nlp_primitives/tests/test_utils.py
```diff
@@ -31,11 +34,8 @@ def test_name_and_desc(self):
         assert any(s in first_word for s in valid_verbs)
         assert self.primitive.input_types is not None
 
-    def test_not_duplicate_of_default(self):
-        class_name = self.primitive.__name__
-        df = list_primitives()
-        primitive_names = df['name'].apply(convert).tolist()
-        assert class_name not in primitive_names
+    def test_name_in_primitive_list(self):
+        assert PRIMITIVES.name.eq(self.primitive.name).any()
```
jeff-hernandez, line 38: When comparing `LSA`, The covert function returns `Lsa`

rwedge (approved):
looks good!

PR #101, opened 2022-02-17
Fixes error with Mean Characters Per Word with Series of Nones

### LINKED ISSUE #100: Mean Characters Per Word returns error on array of None
Mean Characters Per Word returns error on array of None

```python
from nlp_primitives import mean_characters_per_word

mean_characters_per_word = mean_characters_per_word.get_function()

mean_characters_per_word(pd.Series([None, None, None])
```

dvreed77:
Fixes error with Mean Characters Per Word with Series of Nones

- fixes #100

### DIFF
--- a/nlp_primitives/mean_characters_per_word.py
+++ b/nlp_primitives/mean_characters_per_word.py
@@ -5,7 +5,23 @@ import re
 from featuretools.primitives.base import TransformPrimitive
 from woodwork.column_schema import ColumnSchema
 from woodwork.logical_types import Double, NaturalLanguage
+import numpy as np
+import pandas as pd
 
+
+PUNCTUATION = re.escape('!,.:;?')
+END_OF_SENTENCE_PUNCT_RE = re.compile(rf'[{PUNCTUATION}]+$|[{PUNCTUATION}]+ |[{PUNCTUATION}]+\n')
+
+def _mean_characters_per_word(value):
+    if pd.isna(value):
+        return np.nan
+    
+    # replace end-of-sentence punctuation with space
+    value = END_OF_SENTENCE_PUNCT_RE.sub(' ', value)
+    words = value.split()
+    character_count = [len(x) for x in words]
+
+    return np.mean(character_count) if len(character_count) else 0
 
 class MeanCharactersPerWord(TransformPrimitive):
     """Determines the mean number of characters per word.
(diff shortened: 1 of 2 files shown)


dvreed77:
This method also drastically improves performance by over 80x.

CleanShot 2022-02-17 at 10 21 03
CleanShot 2022-02-17 at 10 07 44

nlp_primitives/tests/test_mean_characters_per_word.py
```diff
@@ -41,7 +41,31 @@ def test_nans(self):
                        '',
                        'third line'])
         primitive_func = self.primitive().get_function()
-        answers = pd.Series([np.nan, np.nan, 4.5])
+        answers = pd.Series([np.nan, 0, 4.5])
+        pd.testing.assert_series_equal(primitive_func(x), answers, check_names=False)
+
+    def test_all_nans(self):
```
thehomebrewnerd, line 47: Could we test this with `pd.NA` instead of `np.nan` as a `NaturalLanguage` column with use the pandas `string` dtype which should result in `pd.NA` values being a passed to the primitive function.

Could also maybe parameterize a new test replacing `test_all_nans` and `test_none` to check with `None`, `pd.NA` and `np.nan` as they should all produce the same result.

nlp_primitives/mean_characters_per_word.py
```diff
@@ -29,17 +46,8 @@ class MeanCharactersPerWord(TransformPrimitive):
...
-            x = x.str.replace(end_of_sentence_punct, ' ')
-            # build DF of split words, and calculate avg length
-            df = x.str.split(expand=True)
-            df = df.reset_index()
-            df = df.melt(id_vars='index', var_name='word_index', value_name='word')
-            df['n_characters'] = df['word'].str.len()
-            results = df.groupby(by=['index']).mean(numeric_only=False)['n_characters']
-            return results
+        def mean_characters_per_word(series):
+            series = series.reset_index(drop=True)
```
thehomebrewnerd, line 50: Do we need to do this `reset_index` call?

thehomebrewnerd (commented):
Looks pretty good, just a couple small suggestions/questions. Also should add an entry to `changelog.rst` in the `docs/source` folder.

nlp_primitives/mean_characters_per_word.py
dvreed77: I don't think so. It was already there, and I wasn't sure if it was capturing something I'm not familiar with

thehomebrewnerd (approved):
LGTM, assuming tests pass, of course.

PR #105, opened 2022-02-18
Conditionally add tensorflow dependent primitives to top level primitives

rwedge:
Uses importlib to check for a tensorflow module before adding `UniversalSentenceEncoder`  and `Elmo` to top level imports for the library

### DIFF
--- a/nlp_primitives/tests/test_universal_sentence_encoder.py
+++ b/nlp_primitives/tests/test_universal_sentence_encoder.py
@@ -11,6 +11,7 @@ from featuretools.primitives.utils import (
 from woodwork.logical_types import NaturalLanguage
 
 from ..tensorflow.universal_sentence_encoder import UniversalSentenceEncoder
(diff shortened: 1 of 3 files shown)


nlp_primitives/tests/test_universal_sentence_encoder.py
```diff
@@ -27,6 +28,10 @@ def test_regular(universal_sentence_encoder):
     np.testing.assert_array_almost_equal(a, b)
 
 
+def test_name_in_primitive_list(universal_sentence_encoder):
+    assert PRIMITIVES.name.eq(universal_sentence_encoder.name).any()
```
thehomebrewnerd, line 32: Any reason to add a similar test for `Elmo`?

thehomebrewnerd: Ah, yes. Got it.

thehomebrewnerd (approved):
LGTM!

PR #118, opened 2022-03-30
Initialize LSA pipeline after initialization

rwedge:
Goal of this PR is to move the creation and/or fitting of the LSA pipeline to happen after initialization to avoid a very slow initialization

Also added a `random_seed` parameter with a default value of 0


dvreed77 (approved):
LGTM!

PR #122, opened 2022-04-06
Add serialization test and fix NumUniqueSeparators

### LINKED ISSUE #113: NumUniqueSeparators error when serialized to JSON
Separators is a set which causes error when getting serialized to JSON

thehomebrewnerd:
Moves serialization test from `TestLSA` to `PrimitiveT` so all primitives are tested for ability to serialize.

Fixes serialization issue with `NumUniqueSeparators`

Closes #113

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -6,13 +6,14 @@ Future Release
 ==============
     * Enhancements
     * Fixes
+        * Fix ``NumUniqueSeparators`` to allow for serialization and deserialization (:pr:`122`)
     * Changes
         * Speed up LSA primitive initialization (:pr:`118`)
     * Documentation Changes
     * Testing Changes
 
     Thanks to the following people for contributing to this release:
-    :user:`rwedge`
+    :user:`rwedge`, :user:`thehomebrewnerd`
 
 v2.4.0 Mar 31, 2022
 ===================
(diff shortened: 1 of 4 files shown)


rwedge:
I think the calculate feature matrix calls are extending the total time spent for the unit tests by a lot

Do you think that confirming saving and loading the feature would be sufficient for the test (and skip calculating the feature matrixes)?

thehomebrewnerd:
> Do you think that confirming saving and loading the feature would be sufficient for the test (and skip calculating the feature matrixes)?

I think most of the increased time comes from the Elmo serialization test (at least when I test locally). Testing with CFM might be a little overkill, but if we skip CFM with Elmo I think things will run much faster. Let me test it out.

thehomebrewnerd:
@rwedge Things seem much faster after this change: [e27ee38](https://github.com/alteryx/nlp_primitives/pull/122/commits/e27ee38318bb16d6aec2fd624d640cb9929b800e)

Should we go with something like this or just drop the CFM stuff completely?

rwedge:
@thehomebrewnerd  Let's keep CFM (barring ELMO)

I'd like to make a call to  the serialized feature's primitive `get_function` before serialization, since for `LSA` that's when the  sklearn pipeline is initialized

dvreed77 (approved):
lgtm

PR #127, opened 2022-04-21
Migrate to setup.cfg

### LINKED ISSUE #114: Transition nlp-primitives to use pyproject.toml and setup.cfg (away from setup.py)
> The future of Python packaging is pyproject.toml, and (for now) setup.cfg, based on [PEP 518](https://snarky.ca/clarifying-pep-518/) and (soon) [PEP 621](https://www.python.org/dev/peps/pep-0621/).

- http://ivory.idyll.org/blog/2021-transition-to-pyproject.toml-example.html

- How we did this in Woodwork:
  - https://github.com/alteryx/woodwork/pull/1346
  -  https://github.com/alteryx/woodwork/pull/1361

rwedge:
Closes #114 

Also has more steps in the CI use Make commands instead of using pip directly


.github/workflows/latest_dependency_checker.yml
```diff
@@ -20,18 +20,13 @@ jobs:
       run: |
         python -m virtualenv venv_core
         source venv_core/bin/activate
-        python -m pip install --upgrade pip
-        python -m pip install -r requirements.txt
-        python -m pip install -r test-requirements.txt
```
rwedge, line 25: only dependencies being tracked are `nltk` and `featuretools` which are both regular requirements so not necessary to install the test requirements (which also install the complete requirements)

setup.cfg
```diff
@@ -1,16 +1,79 @@
...
+    pandas >= 1.3.0
+    featuretools >= 1.5.0
+    nltk >= 3.4.5
+    scikit-learn >= 0.20.0
+
+[options.extras_require]
+complete =
+    tensorflow >= 1.14.0; sys_platform!="darwin" or platform_machine!='arm64'
+    tensorflow-metal >= 0.4.0; sys_platform=="darwin" and platform_machine=='arm64'
+    tensorflow-macos >= 2.8.0; sys_platform=="darwin" and platform_machine=='arm64'
```
rwedge, line 43: would be useful to confirm if those constraints are being recognized

thehomebrewnerd: I tested by running `pip install -e ".[complete]" on my M1 mac and got these in my env, so seems to be working.

```
tensorflow-macos==2.8.0
tensorflow-metal==0.4.0
```

nlp_primitives/version.py
```diff
@@ -0,0 +1 @@
+__version__ = '2.5.0'
```
thehomebrewnerd, line 1: Now that we have added this file, should we add a `test_version.py` file like we have in other repos?

Also, we need to update `release.md` to point to the proper places where the version needs to be changed: here and in `test_version.py` if we add that. Don't need to refer to `setup.py` any more.

setup.cfg
rwedge: My non-M1 machine did not install those so seems to be working!

setup.cfg
```diff
@@ -1,16 +1,79 @@
...
+    tensorflow_hub >= 0.4.0
+
+dev =
+    wheel>=0.35.1
```
thehomebrewnerd, line 47: Is `wheel` needed? I know it was in the deps before, but not clear why off the top of my head.

setup.cfg
```diff
@@ -1,16 +1,79 @@
+[metadata]
+name = nlp_primitives
+version = attr: nlp_primitives.version.__version__
+author = Alteryx Inc.
+license = BSD 3-clause
+long_description = file: README.md
+long_description_content_type = text/markdown
+url = https://github.com/alteryx/nlp_primitives
```
thehomebrewnerd, line 8: Maybe we should update the URL's like in Featuretools to include some other links?

https://github.com/alteryx/featuretools/blob/54c2ba8a80193f0b05f271fd493fd7c90fc34f82/setup.cfg#L6

thehomebrewnerd (commented):
We should also update the `make package_nlp_primitives` command to use `python -m build` instead of `python setup.py sdist`. See Featuretools for reference: https://github.com/alteryx/featuretools/blob/54c2ba8a80193f0b05f271fd493fd7c90fc34f82/Makefile#L48

setup.cfg
rwedge: probably not, I imagine `build` would replace it as a direct need at this point if anything

Makefile
```diff
@@ -66,7 +66,7 @@ upgradebuild:
 
 .PHONY: package_nlp_primitives
 package_nlp_primitives: upgradepip upgradebuild
-	python setup.py sdist
-	$(eval PACKAGE=$(shell python setup.py --version))
+	python -m build
+	$(eval PACKAGE=$(shell python -c "from pep517.meta import load; metadata = load('.'); print(metadata.version)"))
```
gsheni, line 70: Is there a reason we are using that versus this? 
```suggestion
	$(eval PACKAGE := $(shell grep '__version__\s=' nlp_primitives/version.py | grep -o '[^ ]*$$'))
```

setup.cfg
gsheni: @thehomebrewnerd I thought `tensorflow-metal` and `tensorflow-macos` are deprecated? Does tensorflow not support M1 Macs natively yet?

gsheni (commented):
left a few comments, but looks good overall.

Makefile
gsheni: Alright

setup.cfg
thehomebrewnerd: As I understand it, there are no pre-built wheels of `tensorflow` that work on M1 macs, so you would have to build it yourself, but I think it would support M1 in that case.

https://github.com/tensorflow/tensorflow/issues/52973

On my M1:
```
>>> pip install tensorflow
ERROR: Could not find a version that satisfies the requirement tensorflow (from versions: none)
ERROR: No matching distribution found for tensorflow
```

gsheni: Gotcha, let's leave this

gsheni:
I can approve this once the tests pass.

gsheni (approved):

PR #132, opened 2022-05-02
Delete setup.py, MANIFEST.in and move configuration to pyproject.toml

gsheni:


.github/workflows/latest_dependency_checker.yml
```diff
@@ -21,14 +21,15 @@ jobs:
       run: |
         python -m virtualenv venv_core
         source venv_core/bin/activate
-        make installdeps
+        python -m pip install ".[test]"
```
thehomebrewnerd, line 24: I don't think we need the test dependencies here. Wouldn't `python -m pip install "."` work as well since we are only checking featuretools and nltk versions?

.github/workflows/latest_dependency_checker.yml
```diff
@@ -21,14 +21,15 @@ jobs:
...
         make checkdeps OUTPUT_FILEPATH=nlp_primitives/tests/requirement_files/latest_core_dependencies.txt
         cat nlp_primitives/tests/requirement_files/latest_core_dependencies.txt
     - name: Update latest complete dependencies
       run: |
         python -m virtualenv venv_complete
         source venv_complete/bin/activate
-        make installdeps-test
+        python -m pip install ".[test]"
```
thehomebrewnerd, line 31: I think we can remove this install command and just rely on `python -m pip install ".[complete]".

gsheni: no, similar to Featuretools, `complete` just installs all add-ons
- https://github.com/alteryx/nlp_primitives/blob/remove_py/setup.cfg#L59

.github/workflows/latest_dependency_checker.yml
gsheni: No, we are checking tensorflow too:
- https://github.com/alteryx/nlp_primitives/blob/remove_py/Makefile#L57

thehomebrewnerd: This is for `make checkdeps` not `make checkdepscomplete` though. And we aren't using the test requirements for anything are we?

.github/workflows/latest_dependency_checker.yml
thehomebrewnerd: I thought `python -m pip install ".[complete]"` installed the base library plus all the add-ons, and we shouldn't need any of the test dependencies for checking the complete requirements.

.github/workflows/latest_dependency_checker.yml
gsheni: Yah actually you're right. Removed this:
- https://github.com/alteryx/nlp_primitives/pull/132/commits/04703a6b80c04833cb31fd08cc165be5cd387aea
- https://github.com/alteryx/nlp_primitives/pull/132/commits/2daee07ec25b9fd07c3984211bed2f6d273d4eaf

.github/workflows/latest_dependency_checker.yml
gsheni: Fixed:
https://github.com/alteryx/nlp_primitives/pull/132/commits/b07b28395842a85243585094e58db41917f39ff7

thehomebrewnerd (dismissed):

thehomebrewnerd (approved):

PR #148, opened 2022-06-13
Allow for custom corpus in LSA primitive

thehomebrewnerd:
Update LSA to allow for users to optionally specify a corpus to use. If not specified, will default to the nltk Gutenberg corpus.


nlp_primitives/lsa.py
```diff
@@ -26,6 +26,11 @@ class LSA(TransformPrimitive):
 
         If a string is missing, return `NaN`.
 
+        Note: If a small custom corpus is used, the output of the primitive may vary
+        depending on the computer architecture being used (linux, macos, windows). This
```
gsheni, line 30:
```suggestion
        depending on the computer architecture being used (Linux, MacOS, Windows). This
```

gsheni (dismissed):
lgtm

rwedge (commented):
should we add a test for a non-randomize algorithm / bad input?

thehomebrewnerd:
> should we add a test for a non-randomize algorithm / bad input?

I can add a quick test for that and code to handle.

rwedge (approved):

PR #154, opened 2022-06-16
Fix bug in CountString with null values

thehomebrewnerd:
Fixes a bug in CountString and related primitives in which a series that had an integer value in the first element but null values further down would result in an error when attempting to calculate feature values. This is due to the fact the the `np.vectorize` function uses the first element to determine the proper dtype to use and would identify `int` which is incompatible with the missing values.

```
ValueError: cannot convert float NaN to integer
```


dvreed77 (approved):

nlp_primitives/count_string.py
```diff
@@ -103,4 +103,4 @@ def count_string(words):
             words = self.process_text(words)
             return len(re.findall(self.pattern, words))
 
-        return np.vectorize(count_string)
+        return np.vectorize(count_string, otypes=[float])
```
rwedge, line 106: Do we have other primitives that are integer nullable return type and float type?  Should we cast to Integer dtype before returning?

thehomebrewnerd: > Do we have other primitives that are integer nullable return type and float type? Should we cast to Integer dtype before returning?

We seem to have a mix of `Double` and `IntegerNullable` return types on our various `Count*` primitives, and I'm not exactly sure why at the moment. I was debating about whether it was important for the actual type of the returned series to match the type specified by `return_type`. The correct type gets set when Woodwork is initialized, so the most critical thing would be for the types to be "compatible" so the conversion could happen.

What is your preference on this?

thehomebrewnerd: I guess I'm also not clear on how we could cast to `Int64` dtype before returning in this case as this primitive is structured differently in that our function `count_string` is just returning a single value `int` or `np.nan`.

rwedge: I think as long as it is compatible with Int64 we can avoid casting

thehomebrewnerd: Yeah, I think float64 to Int64 should always work in this case - especially since we know our numeric values represent integer counts so we don't need to be concerned with loss of information.

rwedge (approved):

PR #161, opened 2022-06-24
Fix LSA naming

thehomebrewnerd:
Overrides the `get_args_string` method to avoid displaying a full user-defined custom corpus in the feature name and instead just displays `corpus=user_defined` instead.

Also updates the algorithm default value to `randomized` instead of `None` to prevent it from always being displayed in feature name.

### DIFF
--- a/nlp_primitives/lsa.py
+++ b/nlp_primitives/lsa.py
@@ -78,13 +78,13 @@ class LSA(TransformPrimitive):
     return_type = ColumnSchema(logical_type=Double, semantic_tags={"numeric"})
     default_value = 0
 
-    def __init__(self, random_seed=0, corpus=None, algorithm=None):
+    def __init__(self, random_seed=0, corpus=None, algorithm="randomized"):
         self.number_output_features = 2
         self.n = 2
         self.trainer = None
         self.random_seed = random_seed
         self.corpus = corpus
-        self.algorithm = algorithm or "randomized"
+        self.algorithm = algorithm
         if self.algorithm not in ["randomized", "arpack"]:
             raise ValueError(
                 "TruncatedSVD algorithm must be either 'randomized' or 'arpack'"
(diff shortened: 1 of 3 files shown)


nlp_primitives/lsa.py
```diff
@@ -78,13 +78,13 @@ class LSA(TransformPrimitive):
...
     default_value = 0
 
-    def __init__(self, random_seed=0, corpus=None, algorithm=None):
+    def __init__(self, random_seed=0, corpus=None, algorithm="randomized"):
         self.number_output_features = 2
         self.n = 2
         self.trainer = None
         self.random_seed = random_seed
         self.corpus = corpus
-        self.algorithm = algorithm or "randomized"
```
dvreed77, line 87: curious if there is a difference doing it this way vs the way it was?

thehomebrewnerd: It impacts how the feature names are generated. The way feature naming works, parameter values are only included in the feature name if the value is changed from the default value. 

With the previous approach the default value was `None` and the algorithm was set to `"randomzied"` inside `__init__` meaning that even though randomized was effectively the default value, it would always be displayed in the resulting feature name because it wasn't the actual default value in the function signature.

Doing it the new way means that `"randomized"` is the default value in the function signature, and the algorithm info will only show up in the feature name if it is changed to the non-default of `"arpack"`.

nlp_primitives/tests/test_lsa.py
```diff
@@ -109,3 +109,17 @@ def test_bad_algorithm_input_value(self):
         err_message = "TruncatedSVD algorithm must be either 'randomized' or 'arpack'"
         with pytest.raises(ValueError, match=err_message):
             LSA(algorithm="bad_algo")
+
+    def test_args_strings(self):
```
rwedge, line 113: can we add another scenario, where random_seed, corpus, and algorithm are all specified

thehomebrewnerd: Added! [bd48d1d](https://github.com/alteryx/nlp_primitives/pull/161/commits/bd48d1d3eaa6cf17327bcd88242ba33e8b3fb54f)

rwedge (approved):
Looks good

PR #170, opened 2022-07-06
Add Woodwork to core dependencies

### LINKED ISSUE #163: Add Woodwork to core dependencies in setup.cfg
Imports from Woodwork are used in several places in the codebase, but Woodwork is not listed as a requirement in `setup.cfg`, but instead we rely on the install of Woodwork as a dependency of Featuretools. Since we directly use Woodwork in the code, we should consider adding it as a core dependency.

sbadithe:
Fixes #163 

-- Adds Woodwork core dependency to nlp_primitives

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -8,11 +8,12 @@ Future Release
     * Fixes
         * Update README.md with Alteryx info (:pr:`167`)
     * Changes
+        * Add Woodwork as core dependency (:pr:`170`)
     * Documentation Changes
     * Testing Changes
 
     Thanks to the following people for contributing to this release:
-    :user:`gsheni`
+    :user:`gsheni`, :user:`sbadithe`
 
 v2.7.1 Jun 29, 2022
 ===================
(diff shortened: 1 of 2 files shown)


sbadithe:
Do we want to support a particular version of Woodwork?

I went ahead and put 0.8.1 for now, because I saw that the featuretools version was >=1.5.0. When I checked out the featuretools 1.5.0 release commit, I saw that woodwork was >=0.8.1 in it's setup.py.

gsheni:
@thehomebrewnerd @sbadithe let's just get this Woodwork dependency listed in setup.cfg. For nlp-primitives, as long as we work with the latest version our library, we will be fine in the short term. Long term, we can handle min dependencies.

sbadithe:
@gsheni  @thehomebrewnerd Sounds good! 

Just writing this down for potential future reference, as I had actually had already played around in the virtualenv: 
I'm not sure if my methodology was correct, but I `pip` uninstalled Pandas, Woodwork, and Featuretools. Then I reinstalled them with `pip` with the minimum versions; what I noticed was `pip` installing featuretools 1.5.0 installed Woodwork 0.16.4 anyways. I uninstalled Woodwork 0.16.4, and specifically installed Woodwork 0.8.1, and tests did pass (with the caveat of the Tensorflow tests as I had to uninstall it to get it to work)

gsheni (approved):

PR #180, opened 2022-07-22
Add NumberOfHashtags and NumberOfMentions primitives

### LINKED ISSUE #139: Add NumberOfHashtags primitive
- Determines the number of hashtags

### LINKED ISSUE #140: Add NumberOfMentions primitive
- Determines the number of mentions (like in a tweet, @_)

sbadithe:
- Fixes #139 
- Fixes #140


nlp_primitives/number_of_hashtags.py
```diff
@@ -0,0 +1,40 @@
...
+
+    name = "number_of_hashtags"
+    input_types = [ColumnSchema(logical_type=NaturalLanguage)]
+    return_type = ColumnSchema(logical_type=Double, semantic_tags={"numeric"})
+    default_value = 0
+
+    def get_function(self):
+        pattern = r"(#[A-Za-z0-9|\_]+)"
+
+        def number_of_hashtags(x):
```
gsheni, line 34: We have a CountString primitive that has similar functionality
- https://github.com/alteryx/nlp_primitives/blob/main/nlp_primitives/count_string.py
We leverage this primitive in other primitives, like WhiteSpaceCount
- https://github.com/alteryx/nlp_primitives/blob/main/nlp_primitives/whitespace_count.py#L6

Should we take the same approach here? Is this a faster implementation (that isn't possible with CountString)?

sbadithe: This primitive would fit well with CountString. One thing is that using CountString would return a np.ndarray but I noticed most of the other primitives return a pd.Series.
I think there are a few other primitives that could be rewritten using CountString as well (maybe more?):

TitleWordCount
UpperCaseCount
PunctuationCount

gsheni: I believe Featuretools can accept either return type. 

Let's create an issue to re-write the other primitives.

sbadithe: Sounds good!

nlp_primitives/number_of_hashtags.py
```diff
@@ -0,0 +1,33 @@
...
+    Description:
+        Given list of strings, determine the number of hashtags
+        in each string. A hashtag is defined as any word starting
+        with a '#' sign followed by a sequence of alphanumeric characters
+        or underscores.
+
+        If a string is missing, return `NaN`.
+
+    Examples:
+        >>> x = ['#regular#expression', 'this is a string', '###__regular#1and_0#expression']
```
thehomebrewnerd, line 20: I guess we can probably debate what a hash tag is exactly, but I believe the `#` symbol that starts a hash tag should be proceeded by a space (or be at the first character position of a the string), and I believe only alpha-numeric characters should be present after the `#` until the space or last character position of the string is reached.

Based on that, I don't think I would call any of these examples valid hashtags.

Did some brief testing and I don't think Twitter recognizes two grouped tags with no space as valid hashtags (`#regular#expression`). It does seem to allow `_` characters though.

My suggestion for the regex would be this:
- `#` symbol must be at first position of the string or preceded by a space
- `#` symbol must be followed by one or more alpha-numeric characters
- Should be terminated by punctuation, a space, or end of string

nlp_primitives/number_of_hashtags.py
```diff
@@ -0,0 +1,33 @@
...
+        >>> number_of_hashtags = NumberOfHashtags()
+        >>> number_of_hashtags(x).tolist()
+        [2.0, 0.0, 3.0]
+    """
+
+    name = "number_of_hashtags"
+    input_types = [ColumnSchema(logical_type=NaturalLanguage)]
+    return_type = ColumnSchema(logical_type=Double, semantic_tags={"numeric"})
```
thehomebrewnerd, line 28: I believe we should use `IntegerNullable` here instead of `Double` since we can never have fractional count values, but we could have missing `nan` values.

nlp_primitives/number_of_mentions.py
```diff
@@ -0,0 +1,33 @@
...
+    Examples:
+         >>> x = ['@portland@oregon', 'this is a string', '@@@__user1@1and_0#expression']
+        >>> number_of_mentions = NumberOfMentions()
+        >>> number_of_mentions(x).tolist()
+        [2.0, 0.0, 2.0]
+    """
+
+    name = "number_of_mentions"
+    input_types = [ColumnSchema(logical_type=NaturalLanguage)]
+    return_type = ColumnSchema(logical_type=Double, semantic_tags={"numeric"})
```
thehomebrewnerd, line 28: `Double` -> `IntegerNullable`

nlp_primitives/number_of_mentions.py
```diff
@@ -0,0 +1,33 @@
...
+    Description:
+        Given list of strings, determine the number of mentions
+        in each string. A mention is defined as any word starting
+        with a '@' sign followed by a sequence of alphanumeric characters
+        or underscores.
+
+        If a string is missing, return `NaN`.
+
+    Examples:
+         >>> x = ['@portland@oregon', 'this is a string', '@@@__user1@1and_0#expression']
```
thehomebrewnerd, line 20: My comment from the hashtags also applies here. I think we should consider updating the regex.

thehomebrewnerd (commented):
I have a few comments about what we consider valid hashtags and mentions. If we make changes to the regex, we'll also need to update the associated tests accordingly.

Aside from some questions about what is a valid hashtag or mention, this PR looks pretty good to me.

gsheni:
@sbadithe What is the status of this? Ready for another review?

sbadithe:
> @sbadithe What is the status of this? Ready for another review?

Still working on this. Trying to get all the edge cases to work (as well as Unicode) by referencing Twitter's open source implementation.

nlp_primitives/number_of_hashtags.py
```diff
@@ -0,0 +1,45 @@
...
+        Given list of strings, determine the number of hashtags
+        in each string.
+
+        A hashtag is defined as a string that meets the following criteria:
+            - Starts at the start of a string or after whitespace
+            - Contains a sequence of alphanumeric characters
+                - The hashtag must contain at least one alphabetic character
+            - Ends with either the end of the string, whitespace, or a non-'#'
+              punctuation character
+                - e.g. #yes-no IS a valid hashtag ("@yes")
```
thehomebrewnerd, line 21: Should the value in parenthesis be `#yes` instead of `@yes`?

nlp_primitives/number_of_mentions.py
```diff
@@ -0,0 +1,49 @@
...
+    Description:
+        Given list of strings, determine the number of mentions
+        in each string.
+
+        A mention is defined as a string that meets the following criteria:
+            - Starts at the start of a string or after whitespace
+            - Contains a sequence of word characters
+            - Ends with either the end of the string, whitespace, or a non-'@'
+              punctuation character
+                - e.g. @yes-no IS a valid mention ("#yes")
```
thehomebrewnerd, line 23: Should the value in parenthesis be `@yes` instead of `#yes`?

nlp_primitives/number_of_hashtags.py
```diff
@@ -0,0 +1,45 @@
...
+    """Determines the number of hashtags in a string.
+
+    Description:
+        Given list of strings, determine the number of hashtags
+        in each string.
+
+        A hashtag is defined as a string that meets the following criteria:
+            - Starts at the start of a string or after whitespace
+            - Contains a sequence of alphanumeric characters
+                - The hashtag must contain at least one alphabetic character
```
thehomebrewnerd, line 18: Combine these: `Contains a sequence of alphanumeric characters with at least one alphabetic character`

nlp_primitives/number_of_hashtags.py
```diff
@@ -0,0 +1,45 @@
...
+            - Ends with either the end of the string, whitespace, or a non-'#'
+              punctuation character
+                - e.g. #yes-no IS a valid hashtag ("@yes")
+                - e.g. #yes# IS NOT a valid hashtag
+
+        This implementation handles Unicode characters.
+
+        This implementation does not implement any sort of character length limit
+        on hashtags.
```
thehomebrewnerd, line 27: Maybe change to `This implementation does not impose any character length limit on hashtags.`

thehomebrewnerd (commented):
Looks pretty good, just a few wording suggestions. Also, the comments that I left on the docstrings apply to both primitives, although I left them in only one place.

thehomebrewnerd (approved):

PR #187, opened 2022-08-02
Add `NumberOfUniqueWords` primitive 

### LINKED ISSUE #138: Add NumberOfUniqueWords primitive
- Determine the n umber of unique words

sbadithe:
- Fixes #138


nlp_primitives/number_of_unique_words.py
```diff
@@ -0,0 +1,62 @@
+import string
```
gsheni, line 1:
```suggestion
from string import punctuation
```

nlp_primitives/number_of_unique_words.py
```diff
@@ -0,0 +1,62 @@
...
+                if pd.isnull(text):
+                    unique_word_cts.append(pd.NA)
+                else:
+                    words = nltk.tokenize.word_tokenize(text)
+                    unique_words = set()
+                    for word in words:
+                        if self.case_insensitive:
+                            word = word.lower().strip(string.punctuation)
+                        else:
+                            word = word.strip(string.punctuation)
```
gsheni, line 56:
```suggestion
                            word = word.strip(punctuation)
```

sbadithe:
Right now this doesn't handle contractions: 

`can't` is tokenized by `nltk` into `can` and `t`. I'm working to see if there is a way around that

docs/source/changelog.rst
```diff
@@ -5,7 +5,7 @@ Changelog
 Future Release
 ==============
     * Enhancements
-        * Add NumberOfHashtags and NumberOfMentions primitives (:pr:`180`)
+        * Add `NumberOfHashtags`, `NumberOfUniqueWords` and `NumberOfMentions`` primitives (:pr:`180`) (:pr:`187`)
```
thehomebrewnerd, line 8: I'd add this PR as as separate enhancement line item, since it is unrelated to the NumberOfHashtags and NubmerOfMentions work.

nlp_primitives/number_of_unique_words.py
```diff
@@ -0,0 +1,57 @@
...
+
+class NumberOfUniqueWords(TransformPrimitive):
+    """Determines the number of unique words in a string.
+
+    Description:
+        Determines the number of unique words in a given string. Includes options for
+        case-insensitive behavior.
+
+    Args:
+        case_insensitive (bool): Specify case_insensitivity when searching for unique words.
```
thehomebrewnerd, line 17: Normally for optional arguments we list the type a `(bool, optional)` and then mention the default value at the end of the description, so someone reading this clearly knows what the default is, without having to look at the code. "Defaults to False".

nlp_primitives/number_of_unique_words.py
```diff
@@ -0,0 +1,57 @@
...
+
+    default_value = 0
+
+    def __init__(self, case_insensitive=False):
+        self.case_insensitive = case_insensitive
+
+    def get_function(self):
+        def num_unique_words(array):
+            unique_word_cts = []
+            for text in array:
```
thehomebrewnerd, line 45: I'm a little concerned about the performance of this implementation. Normally we want to try to avoid looping over the rows in a series if at all possible. Is there a way this can be vectorized? If not, is it better to create a counting function and then use `.apply` to create the new series of counts?

I think if nothing else you should move the case insensitive check outside of the for loop and do that as a vectorized operation with `array.str.lower()` instead of doing it on each row/word individually. Right now you do the `if self.case_insensitive` check every time through the `for` loop, but you really only need to do that check one time at the start.

thehomebrewnerd (commented):
How does this handle numbers that are present in a string, or things like email addresses or URLs?

Examples:
-user123
-$34.95
-user@email.com
-google.com
-https://alteryx.com

nlp_primitives/number_of_unique_words.py
thehomebrewnerd: Another thought on this - can we use the logic from the `NumberOfCommonWords` primitive to determine how many words are present, rather than creating different logic for determining what is a word in this primitive?

https://github.com/alteryx/nlp_primitives/blob/399cfc2273d87bf750648ea196336b4bb387c22f/nlp_primitives/number_of_common_words.py#L61

sbadithe: @thehomebrewnerd  @gsheni 
Thanks for the review. I refactored the code to use vectorized `pandas.str.split`. 

However, I did have some questions regarding intended behavior: 
-$34.95
[-user@email.com](mailto:-user@email.com)
-google.com
-https://alteryx.com/

I removed the `.` character as a character to split on. This keeps `alteryx.com` as a single string. However I ran into two issues: 
(1) handling the ellipsis ("this...this"). I've had some trouble getting the regex to match only 3 periods (`\.{3}` doesn't seem to be working).  
(2) `NumberOfCommonWords` seems to behave differently. It splits on `.` so that `alteryx.com` becomes 2 words. Thus if we use the input string `I have $0.99. visit alteryx.com` and pass in the common words `["$0.99", "alteryx.com"]`, we will get 0 matches. (Passing in `["alteryx", "99"]` gets two matches). I agree with you though, as a user, I would not expect those to be split. 

Additionally,  stripping the strings of punctuation becomes necessary: Given input `test test?` I would want `NumberOfUniqueWords` to return 1. This does cause a bit of a perf hit though.

gsheni: - How much of a performance hit is there?
If you think some of these are edge cases, we can punt on them, and create follow up issues.

sbadithe: From cursory testing, on an input size of 50,000, adding these checks adds about ~45ms. For reference, it is ~20ms slower than `NumberOfCommonWords`. Given we are computing more, I don't feel this to be a huge issue. I also think that a lot of these checks are unavoidable in order to implement reasonable behavior (A user would expect `test test.` to return 1 in my opinion). Although, certainly, I can test the perf implications more thoroughly.   

What gives me pause is that the semantics are different from `NumberOfCommonWords`. A user using that primitive would have `alteryx.com` split into two words. `$9.99` would be split into two.  

I think our primitives should ideally have uniform behavior for these cases so that users can be confident about their use. I.e. a user who wanted to know how many unique words were in `visit alteryx.com` would probably expect 2, and 3 would be a surprising answer. 

My personal feeling is we should probably maintain this behavior and file an issue to revisit what we want the desired behavior of `NumberOfCommonWords` to be.  

On further research, regarding the ellipsis edge case, I think we are fine (if not fully correct) to ignore it. Chicago Manual of Style indicates there should be a space after the ellipsis so I think this means we are fine.

thehomebrewnerd: One more thing to think about - do words contain numbers and symbols or only letters? Are `alteryx.com` and `$9.99` words?

ozzieD: I really think we should get the opinion of some of the DS folks. But imo we should be trying to avoid forcing a user toward one way of tokenizing words. NLTK has a multitude of ways of doing this because NLP is highly dependent on what a user wants to do. 

https://stackoverflow.com/questions/34714162/preventing-splitting-at-apostrophies-when-tokenizing-words-using-nltk

I think we should stick with a defualt tokenizer with these primitives and allow the user to change this defualt if they please.

nlp_primitives/number_of_unique_words.py
```diff
@@ -0,0 +1,62 @@
...
+    return_type = ColumnSchema(logical_type=IntegerNullable, semantic_tags={"numeric"})
+
+    default_value = 0
+
+    def __init__(self, case_insensitive=False):
+        self.case_insensitive = case_insensitive
+
+    def get_function(self):
+        def _unique_word_helper(text):
+            if not isinstance(text, Iterable):
```
ozzieD, line 46: can check for `str` instead of `Iterable` and save yourself the import

nlp_primitives/number_of_words_in_quotes.py
```diff
@@ -0,0 +1,47 @@
...
+    """
+
+    name = "number_of_words_in_quotes"
+    input_types = [ColumnSchema(logical_type=NaturalLanguage)]
+    return_type = ColumnSchema(logical_type=IntegerNullable, semantic_tags={"numeric"})
+    default_value = 0
+
+    def __init__(self, double_only=False, single_only=False):
+        if double_only and single_only:
+            raise Exception
```
ozzieD, line 38: Not sure if I like the two parameter initialization. This could be a one parameter called `quote_type` with inputs being either `["double", "single", "both"]`.

nlp_primitives/number_of_unique_words.py
sbadithe: After discussion, the decision was made to mimic `NLTK`'s behavior for the default behavior. 
For example: 
-- `alteryx.com` will be tokenized as `alteryx.com`
-- `$3.99` will be tokenized as `$` and `3.99`
-- `half-baked` will be tokenized as `half-baked` 
-- `test*` will be tokenized as `test`

nlp_primitives/number_of_unique_words.py
sbadithe: Hmm, maybe I misunderstood something, but this didn't seem to work

nlp_primitives/number_of_words_in_quotes.py
sbadithe: Good point, I think I agree. I've separated `NumberOfWordsInQuotes` out into another PR for now.

nlp_primitives/tests/test_number_of_unique_words.py
```diff
@@ -0,0 +1,84 @@
...
+        x = pd.Series(
+            [
+                "test test test test",
+                "test TEST test TEST",
+                "and;subsequent;lines...",
+                "$0.99 alteryx@alteryx.com",
+            ]
+        )
+
+        expected = pd.Series([1, 2, 3, 2])
```
ozzieD, line 21: Shouldn't the expected for `test TEST test TEST` be 1?

sbadithe: Default behavior is for `case_insensitive` to be `False`

ozzieD: Got it!

ozzieD (approved):
LGTM

PR #188, opened 2022-08-02
Implement `MeanCharactersPerSentence` and `NumberOfSentences` primitives

### LINKED ISSUE #137: Add NumberOfSentences primitive
- Determine the number of sentences

### LINKED ISSUE #141: Add MeanSentenceLength primitive
- Determines the average length of the sentences

sbadithe:
- Fixes  #141 
- Fixes  #137


sbadithe:
The `MeanSentenceLength` primitive determines the mean length in characters. It includes closing punctuation.

Some thoughts

-- Should we include closing punctuation? 
-- Should we change the primitive name to make it clearer that it is determining length in terms of characters, and not number of words? Perhaps we could add a `MeanSentenceLengthInWords` primitive later?

nlp_primitives/mean_sentence_length.py
```diff
@@ -0,0 +1,45 @@
...
+
+    name = "mean_length_of_sentences"
+    input_types = [ColumnSchema(logical_type=NaturalLanguage)]
+    return_type = ColumnSchema(logical_type=Double, semantic_tags={"numeric"})
+    default_value = 0
+
+    def get_function(self):
+        def helper(text):
+            if not isinstance(text, Iterable):
+                return np.nan
```
ozzieD, line 35: This works but instead of having to import `Iterable` you could just check for the instance of a `str` since the `NaturalLanguage` logical type is always a str physical type.

nlp_primitives/mean_sentence_length.py
```diff
@@ -0,0 +1,45 @@
...
+        [5.0, 4.0, 6.0]
+    """
+
+    name = "mean_length_of_sentences"
+    input_types = [ColumnSchema(logical_type=NaturalLanguage)]
+    return_type = ColumnSchema(logical_type=Double, semantic_tags={"numeric"})
+    default_value = 0
+
+    def get_function(self):
+        def helper(text):
```
ozzieD, line 33: Nitpick I'm not a fan of just a function just called `helper` even if it is in a simple context like this. I would call this something like `_mean_len_of_sent` or something like that.

nlp_primitives/number_of_sentences.py
```diff
@@ -0,0 +1,44 @@
...
+
+    name = "number_of_sentences"
+    input_types = [ColumnSchema(logical_type=NaturalLanguage)]
+    return_type = ColumnSchema(logical_type=IntegerNullable, semantic_tags={"numeric"})
+    default_value = 0
+
+    def get_function(self):
+        def helper(text):
+            if not isinstance(text, Iterable):
+                return np.nan
```
ozzieD, line 35: same comments from the other primitive apply here

nlp_primitives/mean_sentence_length.py
sbadithe: I don't think this works because the input to the helper function is a `list`.

nlp_primitives/mean_characters_per_sentence.py
```diff
@@ -0,0 +1,47 @@
...
+    return_type = ColumnSchema(logical_type=Double, semantic_tags={"numeric"})
+    default_value = 0
+
+    def get_function(self):
+        def _mean_characters_per_sentence(text):
+            if not isinstance(text, Iterable):
+                return np.nan
+            if len(text) == 0:
+                return 0
+            sentences = sent_tokenize(text)
```
sbadithe, line 38: On an input series with 50,000 elements, each a string of length ~4300 characters, this approach yielded a ~20% performance improvement in time versus  `np.mean([len(s) for s in sentences])`

nlp_primitives/mean_characters_per_sentence.py
```diff
@@ -0,0 +1,47 @@
...
+            total = 0.0
```
sbadithe, line 39: Tested on input with 50,000 elements in a series, each consisting of a string of length ~4300. Had ~20% perf improvement over using `np.mean([len(s) for s in sentences])`

docs/source/changelog.rst
```diff
@@ -6,6 +6,7 @@ Future Release
 ==============
     * Enhancements
         * Add NumberOfHashtags and NumberOfMentions primitives (:pr:`180`)
```
gsheni, line 8:
```suggestion
        * Add `NumberOfHashtags` and `NumberOfMentions` primitives (:pr:`180`)
```

nlp_primitives/mean_characters_per_sentence.py
```diff
@@ -0,0 +1,47 @@
...
+
+    name = "mean_characters_per_sentence"
+    input_types = [ColumnSchema(logical_type=NaturalLanguage)]
+    return_type = ColumnSchema(logical_type=Double, semantic_tags={"numeric"})
+    default_value = 0
+
+    def get_function(self):
+        def _mean_characters_per_sentence(text):
+            if not isinstance(text, Iterable):
+                return np.nan
```
ozzieD, line 35: This import you can save because what comes in here is a string

nlp_primitives/number_of_sentences.py
```diff
@@ -0,0 +1,44 @@
...
+    """
+
+    name = "number_of_sentences"
+    input_types = [ColumnSchema(logical_type=NaturalLanguage)]
+    return_type = ColumnSchema(logical_type=IntegerNullable, semantic_tags={"numeric"})
+    default_value = 0
+
+    def get_function(self):
+        def _number_of_sentences(text):
+            if not isinstance(text, Iterable):
```
ozzieD, line 34: same with this one

ozzieD (approved):

PR #193, opened 2022-08-10
Remove readthedocs and docs requirements

gsheni:
- We don't use readthedocs in this repo so there isn't any point in tracking doc requirements
- Will prevent PRs like this: https://github.com/alteryx/nlp_primitives/pull/192


rwedge:
Should we delete the docs folder and have a changelog.rst / changelog.md in the root directory?

gsheni:
@rwedge good point. Fixed

sbadithe (commented):
Should we remove the docs from `.gitignore` as well?

sbadithe (approved):

PR #196, opened 2022-08-12
Add `NumberOfWordsInQuotes` Primitive

### LINKED ISSUE #136: Add NumberOfWordsInQuotes primitive
- Determine the number of words in single/double quotes

sbadithe:
- Fixes #136


nlp_primitives/number_of_words_in_quotes.py
```diff
@@ -0,0 +1,69 @@
...
+
+    Description:
+        Given a list of strings, determine the number of words in quotes
+        in each string.
+
+        This implementation handles Unicode characters.
+
+        If a string is missing, return `NaN`.
+
+    Examples:
```
gsheni, line 22: We need to add an args section

nlp_primitives/number_of_words_in_quotes.py
```diff
@@ -0,0 +1,69 @@
...
+    input_types = [ColumnSchema(logical_type=NaturalLanguage)]
+    return_type = ColumnSchema(logical_type=IntegerNullable, semantic_tags={"numeric"})
+    default_value = 0
+
+    def __init__(self, capture=None):
+        self.capture = "both"
+        if capture == "single":
+            self.capture = "single"
+        elif capture == "double":
+            self.capture = "double"
```
gsheni, line 39:
```suggestion
    def __init__(self, capture=None):
        if capture not in ['both', 'single', 'double']:
            raise ValueError("Invalid capture")
        self.capture = capture
```

nlp_primitives/number_of_words_in_quotes.py
gsheni: Also, why capture of `None`? What does that mean? Our primitives should work if the argument is not specified.

sbadithe: I should set default value to `both` just to be clear.  Currently it still works -- just would be set to default in the actual `get_function` logic, but I agree that that is unclear.

gsheni: yah, default to `both`

nlp_primitives/tests/test_number_of_words_in_quotes.py
```diff
@@ -0,0 +1,84 @@
...
+            [
+                'Yes "    "',
+                '"Hello this is a test"',
+                '"Yes" "   "',
+                "",
+                '"Python, java prolog"',
+                '"Python, java prolog" three words here "binary search algorithm"',
+            ]
+        )
+        expected = pd.Series([0, 5, 1, 0, 3, 6]).astype("Int64")
```
thehomebrewnerd, line 23: For these expected calls you can specify the dtype directly when creating the series without having to chain on the `.astype` casting:
```
pd.Series([0, 5, 1, 0, 3, 6], dtype="Int64")
```

nlp_primitives/tests/test_number_of_words_in_quotes.py
```diff
@@ -0,0 +1,84 @@
...
+            [
+                "'Yes\n, this is me'",
+            ]
+        )
+        expected = pd.Series([4.0]).astype("Int64")
+        actual = self.primitive().get_function()(x)
+        pd.testing.assert_series_equal(actual, expected, check_names=False)
+
+    def test_raises_error_invalid_args(self):
+        with pytest.raises(ValueError):
```
thehomebrewnerd, line 71: Also would be good to check that the error message matches the expected match, that way you know you aren't creating some other unrelated `ValueError`.
```
error_msg = "expected error message text"
with pytest.raises(ValueError, match=error_msg):
    ...
```

nlp_primitives/tests/test_number_of_words_in_quotes.py
```diff
@@ -0,0 +1,84 @@
...
+            self.primitive(capture="NULL")
+
+    def test_null(self):
+        x = pd.Series([np.nan, pd.NA, None, '"test"'])
+        actual = self.primitive().get_function()(x)
+        expected = pd.Series([np.nan, np.nan, np.nan, 1.0]).astype("Int64")
```
thehomebrewnerd, line 77: If we are using `Int64` the `np.nan` values will get converted to `pd.NA` values, so you could just specify them as that here.

thehomebrewnerd (commented):
The implementation feels a little complicated to me, but I'm not sure how to clean it up off the top of my head (maybe we can't), but I'd like to think about that just a bit more.

nlp_primitives/number_of_words_in_quotes.py
```diff
@@ -0,0 +1,76 @@
...
+        def num_words_in_quotes(array):
+            IN_DOUBLE_QUOTES = r'"([^"]+)"'
+            IN_SINGLE_QUOTES = r"'([^']+)'"
+            if self.capture == "single":
+                text = array.str.extractall(IN_SINGLE_QUOTES)
+            elif self.capture == "double":
+                text = array.str.extractall(IN_DOUBLE_QUOTES)
+            else:
+                regex = f"({IN_SINGLE_QUOTES}|{IN_DOUBLE_QUOTES})"
+                text = array.str.extractall(f"{regex}")
```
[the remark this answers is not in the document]
sbadithe: Thanks, I think the above looks a lot better.

nlp_primitives/number_of_words_in_quotes.py
```diff
@@ -0,0 +1,76 @@
...
+        self.capture = capture
+
+    def get_function(self):
+        def _word_tokenize(text):
+            tokens = word_tokenize(text)
+            ct = 0
+            for word in tokens:
+                if len(word.strip(punctuation).strip()) > 0:
+                    ct += 1
+            return ct
```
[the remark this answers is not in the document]
sbadithe: I agree that the comprehension is neater, but I'm concerned about the performance implications of allocating that much continuous memory. On `NumberOfUniqueWords`, I had a 20% perf improvement on large strings by replacing a comprehension with iteration and a `ct` variable.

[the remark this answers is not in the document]
thehomebrewnerd: That's fair.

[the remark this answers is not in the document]
thehomebrewnerd: What about just `return sum(len(word.strip(punctuation + " ")) > 0 for word in word_tokenize(text))`? I think that might use a generator internally instead of allocating memory for a list. Not sure how that performance would compare...or if that makes it less readable.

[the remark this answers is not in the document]
sbadithe: Was just thinking about a similar thing! Will try it out now

sbadithe:
> The implementation feels a little complicated to me, but I'm not sure how to clean it up off the top of my head (maybe we can't), but I'd like to think about that just a bit more.

I agree that it does seem complex. I think I found the main complexity to be because of the following: 

-- Strings can have several matches, all of which need to be tokenized. Then we need to add the summation of all matches to the same value in the series.
-- Strings can have no matches --> `extractall` just omits them in this case, and we need to `reindex` to "get" those strings "back".
-- Can get a pd.NA 

I'm not experienced in Pandas by any means, so I wouldn't be surprised if there was a better way to do this!

thehomebrewnerd:
> I agree that it does seem likely a little too complex. I think I found the main complexity to be because of the following:
> 
> -- Strings can have several matches, all of which need to be tokenized. Then we need to add the summation of all matches to the same index in the series. -- Strings can have no matches --> `extractall` just omits them in this case, and we need to `reindex` to "get" those strings "back" (and set the matches to 0). -- Can get a pd.NA

I haven't completely thought this through, and this doesn't work quite right for the "both" capture case, but did you explore an implementation somewhat like this?
```python
    def __init__(self, capture="both"):
        if capture not in ["both", "single", "double"]:
            raise ValueError(
                f"{capture} is not a valid argument. Specify 'both', 'single', or 'double'"
            )
        self.capture = capture
        IN_DOUBLE_QUOTES = r'"([^"]+)"'
        IN_SINGLE_QUOTES = r"'([^']+)'"
        if capture == "double":
            self.regex = IN_DOUBLE_QUOTES
        elif capture == "single":
            self.regex = IN_SINGLE_QUOTES
        else:
            self.regex = f"({IN_SINGLE_QUOTES}|{IN_DOUBLE_QUOTES})"


    def get_function(self):
        def count_words_in_quotes(text):
            if pd.isnull(text):
                return pd.NA
            matches = re.findall(self.regex, text)
            if matches and isinstance(matches[0], tuple):
                matches = [match for group in matches for match in group]
            return sum(len(match.split()) for match in matches)

        def num_words_in_quotes(array):
            return array.apply(count_words_in_quotes).astype("Int64")
        return num_words_in_quotes
```

I guess I wasn't clear what we were gaining by stripping punctuation and whitespace and using nltk to tokenize, instead of just splitting the matches on whitespace characters. I might not be thinking of some edge cases though.

sbadithe:
> 

Thanks! I will definitely try this out. I think this should work on the `both` case if we strip out the grouping in the regexes!

Regarding using `nltk` tokenize, I was just using that for convenience, and in an attempt to try and make sure edge cases lined up across primitives. Looking at the source code, `NLTK` uses a pretty simple regex that I think we can apply directly. I agree that we can just remove the function call. (As an aside, I wonder if it is worth exploring defining some standard regexes at the module level so that we can standardize across primitives)

Regarding the need to call `strip`, I believe it's necessary to ensure the input string "'......'" returns 0.

Regarding the use of `apply` vs `extractall`, I originally reached for `extractall` because the `Pandas` documentation seemed to indicate that `.str` functions were vectorized. I didn't actually profile the results, and now looking at the `Pandas` source code, I'm not sure if it is vectorized in the sense I thought it was. I will invest some time learning to profile, and get back to you on any potential performance differences...

thehomebrewnerd:
> Regarding the use of `apply` vs `extractall`, I originally reached for `extractall` because the `Pandas` documentation seemed to indicate that `.str` functions were vectorized. I didn't actually profile the results, and now looking at the `Pandas` source code, I'm not sure if it is vectorized in the sense I thought it was. I will invest some time learning to profile, and get back to you on any potential performance differences.

Yeah, might be good to study a little more. In the original implementation you end up doing an `.apply` call after `extractall`, so maybe it's better to just wrap all those operations in a single apply call? Will be interesting to see what you find out from profling.

sbadithe:
> > Regarding the use of `apply` vs `extractall`, I originally reached for `extractall` because the `Pandas` documentation seemed to indicate that `.str` functions were vectorized. I didn't actually profile the results, and now looking at the `Pandas` source code, I'm not sure if it is vectorized in the sense I thought it was. I will invest some time learning to profile, and get back to you on any potential performance differences.
> 
> Yeah, might be good to study a little more. In the original implementation you end up doing an `.apply` call after `extractall`, so maybe it's better to just wrap all those operations in a single apply call? Will be interesting to see what you find out from profling.

Here were the results on an input series of length 5000, each with an input string of length 15,000:

The original (calling `word_tokenize` from NLTK): 183.39 seconds
The original, but replacing `word_tokenize` with local regex: 8.971 seconds
The above suggested: 3.792 seconds 
The above suggested, without comprehension: 2.56 seconds

Takeaways: I guess `.str.extractall`, even if it is vectorized, is costly. Additionally, in the future we should probably stay away from `nltk` functions if we can use our own regex, as tokenizing seems extremely costly.

sbadithe:
> > Regarding the use of `apply` vs `extractall`, I originally reached for `extractall` because the `Pandas` documentation seemed to indicate that `.str` functions were vectorized. I didn't actually profile the results, and now looking at the `Pandas` source code, I'm not sure if it is vectorized in the sense I thought it was. I will invest some time learning to profile, and get back to you on any potential performance differences.
> 
> Yeah, might be good to study a little more. In the original implementation you end up doing an `.apply` call after `extractall`, so maybe it's better to just wrap all those operations in a single apply call? Will be interesting to see what you find out from profling.

Here were the results on an input series of length 5000, each with an input string of length 15,000: 
The original (calling `word_tokenize` from `NLTK`): 183.39 seconds 
The original, but replacing `word_tokenize` with local regex: 8.971 seconds 
The above suggested: 3.792 seconds 
The above suggested, without comprehension: 2.56 seconds 

I would conjecture the perf change is because `word_tokenize` stores all words in memory in a list, but I am not sure. I guess `.str.extractall`, even if it is vectorized, is costly (I think here the meaning of `vectorized` is that we can do batch operations, not necessarily that it is in C). Additionally, in the future we should probably stay away from `nltk` functions if we can use our own regex, as tokenizing seems extremely costly.

sbadithe:
I suppose there is an edge case where the user selects `both` or `single` quote_type and there are contractions in the corpus. In that case, we are splitting the word into two. I'm not sure of the best behavior for this case.

thehomebrewnerd:
> I suppose there is an edge case where the user selects `both` or `single` quote_type and there are contractions in the corpus. In that case, we are splitting the word into two. I'm not sure of the most accurate and meaningful behavior for this case (if there even is one).

I think we should include examples in our tests of various "complicated" words to make sure things behave as we expect. For example - email address, url, hyphenated word, contractions, etc.

Also, thanks for doing the performance testing!

sbadithe:
Thanks! I changed the regex to use lazy quantifiers and I think contractions should now work. I also added some special cases to the tests.

thehomebrewnerd (approved):
@sbadithe This looks good! Thanks for your work and investigation on this to arrive at a good solution!

Two more questions that I don't consider blockers for merging:
- This doesn't have any problem with strings that contain a single quote, correct? Something like `'There should be " 0 words in quotes in this string'`
- How are numbers handled? Are `"$9.95"` and `"100"` and `"100,000"` all considered one word or are they not considered as words? We might want to clarify that in the doctsring if the behavior isn't obvious.

sbadithe:
> @sbadithe This looks good! Thanks for your work and investigation on this to arrive at a good solution!
> 
> Two more questions that I don't consider blockers for merging:
> 
> * This doesn't have any problem with strings that contain a single quote, correct? Something like `'There should be " 0 words in quotes in this string'`
> * How are numbers handled? Are `"$9.95"` and `"100"` and `"100,000"` all considered one word or are they not considered as words? We might want to clarify that in the doctsring if the behavior isn't obvious.

1) No, there shouldn't be any issues. As far as I have tested, it can handle contractions and stray marks fine. I also just tested it on escaped quotations (`there's a \'stray\' quote`) and it correctly identified one quoted word. 
2) It doesn't split on `.` but it does split on `,`, so `34.99` is one word but `100,000` is two.  I think `100,000` should stay as one. This is NLTK's behavior too.
I can remove `,` as a delimiter to keep `100,000` as 1 word. 

Looking at the other primitives, like `NumberOfCommonWords` , we do delimit on commas, so this behavior would be a little different. Should we, in a future PR, pull the delimiter regex out of the hands of each primitive and define it on a module-basis to iron out any discrepancies?

thehomebrewnerd:
> Looking at the other primitives, like `NumberOfCommonWords` , we do delimit on commas, so this behavior would be a little different. Should we, in a future PR, pull the delimiter regex out of the hands of each primitive and define it on a module-basis to iron out any discrepancies?

I think it makes sense to not split on commas. The downside to that would be that for a sentence that didn't include a space after a comma (`word,word`), we wouldn't identify those as two words. Not sure, but it seems more common to have commas present in numbers than to not have a space after a comma in a normal sentence.

I also think it would be good to clean up the delimiters and use a common set across all primitives, if possible. Could you create an issue for that?

sbadithe:
> > Looking at the other primitives, like `NumberOfCommonWords` , we do delimit on commas, so this behavior would be a little different. Should we, in a future PR, pull the delimiter regex out of the hands of each primitive and define it on a module-basis to iron out any discrepancies?
> 
> I think it makes sense to not split on commas. The downside to that would be that for a sentence that didn't include a space after a comma (`word,word`), we wouldn't identify those as two words. Not sure, but it seems more common to have commas present in numbers than to not have a space after a comma in a normal sentence.
> 
> I also think it would be good to clean up the delimiters and use a common set across all primitives, if possible. Could you create an issue for that?

Sounds good!

thehomebrewnerd (approved):

PR #201, opened 2022-08-29
Use pyproject.toml only (move away from setup.cfg) 

gsheni:


nlp_primitives/__init__.py
```diff
@@ -1,36 +1,48 @@
...
+    if (
+        inspect.isclass(obj)
+        and obj is not AggregationPrimitive
+        and obj is not TransformPrimitive
+        and issubclass(obj, (AggregationPrimitive, TransformPrimitive))
+    )
+]
 
-nltk_data_path = pkg_resources.resource_filename("nlp_primitives", "data/nltk-data/")
+nltk_data_path = pkg_resources.resource_filename("nlp_primitives", "data/nltk_data/")
```
gsheni, line 47: Renamed this to underscore, since its more correct (Python wise)

nlp_primitives/required_data.txt
```diff
@@ -1 +1 @@
-nltk-data.tar.gz
\ No newline at end of file
+nltk_data.tar.gz
```
gsheni, line 1: @rwedge @thehomebrewnerd do we actually need this?

thehomebrewnerd: Would have to study a bit more to know for sure, but it looks like we are including the data we need by including the `data/nltk_data` directory, so it doesn't seem like that file is needed for anything.

gsheni: Removed

.flake8
```diff
@@ -0,0 +1,8 @@
+[flake8]
+exclude = docs/*
+max-line-length = 88
+extend-ignore = E203
+ignore = E501,W504,W503
+per-file-ignores =
+    **/__init__.py:F401
+    **/tests/*:D
```
rwedge, line 8: what is the D class of errors?

gsheni: Removed:
https://github.com/alteryx/nlp_primitives/pull/201/commits/d5a5f3114685cdf36b79e258b4241c6c8b5486d2

rwedge (approved):

PR #211, opened 2022-10-12
Fix case-sensitive behavior in `clean_tokens`

### LINKED ISSUE #210: `clean_tokens` performs case sensitive comparison to `nltk.corpus.stopwords`
For example, "yourself" is a stopword but "Yourself" is not being matched as a stopword. I think this is a little strange, because our `StopwordCount` primitive is case insensitive. One possible behavior we can investigate implementing is to call `.lower()` on the word before checking whether it is a stopword.

sbadithe:
- Fixes #210


nlp_primitives/utilities.py
```diff
@@ -1,19 +1,23 @@
...
 
 import nltk
 
 
-def clean_tokens(textstr):
-    wn = nltk.WordNetLemmatizer()
-
-    textstr = textstr.translate(str.maketrans("", "", string.punctuation))
-    textstr = [ch for ch in textstr.split(" ") if len(ch) > 0]
+def clean_tokens(text: str) -> List[str]:
```
sbadithe, line 7: took this opportunity to refactor this function. happy to revert changes if anything is worse now!

thehomebrewnerd (approved):
LGTM!

PR #234, opened 2022-12-20
Fix handling of all-whitespace strings in `MeanCharactersPerSentence`

### LINKED ISSUE #232: MeanCharactersPerSentence fails on just whitespace text
When there is text that is only whitespace, this text passes through the edge case checks in `_mean_characters_per_sentence` and `len(sentences)` ends up being zero which causes a divide my zero error. 

```
Exception has occurred: ZeroDivisionError       (note: full exception trace is shown but execution is paused at: _run_module_as_main)
float division by zero
  File "/Users/odembowski/Repos/nlp_primitives/nlp_primitives/mean_characters_per_sentence.py", line 42, in _mean_characters_per_sentence
    return total / len(sentences)
  File "/Users/odembowski/.pyenv/versions/feature_tools/lib/python3.8/site-packages/pandas/_libs/lib.pyx", line 2870, in pandas._libs.lib.map_infer
    val = f(arr[i])
  File "/Users/odembowski/.pyenv/versions/feature_tools/lib/python3.8/site-packages/pandas/core/apply.py", line 1143, in apply_standard
    mapped = lib.map_infer(
  File "/Users/odembowski/.pyenv/versions/feature_tools/lib/python3.8/site-packages/pandas/core/apply.py", line 1088, in apply
    return self.apply_standard()
  File "/Users/odembowski/.pyenv/versions/feature_tools/lib/python3.8/site-packages/pandas/core/series.py", line 4433, in apply
    return SeriesApply(self, func, convert_dtype, args, kwargs).apply()
  File "/Users/odembowski/Repos/nlp_primitives/nlp_primitives/mean_characters_per_sentence.py", line 45, in mean_characters_per_sentence
    return array.apply(_mean_characters_per_sentence)
  File "/Users/odembowski/Repos/featuretools/rec_prims.py", line 45, in <module>
    answer = get_fun(es["df"]["text"])
  File "/Users/odembowski/.pyenv/versions/3.8.12_x86/lib/python3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/Users/odembowski/.pyenv/versions/3.8.12_x86/lib/python3.8/runpy.py", line 194, in _run_module_as_main (Current frame)
    return _run_code(code, main_globals, None,
```
along with these checks 
```
            if not isinstance(text, Iterable):
                return np.nan
            if len(text) == 0:
                return 0
```
we may need to check for just whitespace and return 0 or nan. Or return 0 if the `sent_tokenize` tokenizer doesn't return anything.

sbadithe:
- closes #232

### DIFF
--- a/nlp_primitives/mean_characters_per_sentence.py
+++ b/nlp_primitives/mean_characters_per_sentence.py
@@ -33,9 +33,7 @@ class MeanCharactersPerSentence(TransformPrimitive):
                 return np.nan
             if not len(sentences := sent_tokenize(text)):
                 return 0.0
-            total = 0.0
-            for s in sentences:
-                total += len(s)
+            total = sum(len(s) for s in sentences)
             return total / len(sentences)
 
         def mean_characters_per_sentence(array):
--- a/nlp_primitives/mean_characters_per_sentence.py
+++ b/nlp_primitives/mean_characters_per_sentence.py
@@ -31,9 +31,7 @@ class MeanCharactersPerSentence(TransformPrimitive):
         def _mean_characters_per_sentence(text):
             if not isinstance(text, str):
                 return np.nan
-            if len(text) == 0:
-                return 0.0
-            if not sentences := sent_tokenize(text):
+            if not len(sentences := sent_tokenize(text)):
                 return 0.0
             total = 0.0
             for s in sentences:
--- a/nlp_primitives/mean_characters_per_sentence.py
+++ b/nlp_primitives/mean_characters_per_sentence.py
@@ -32,8 +32,9 @@ class MeanCharactersPerSentence(TransformPrimitive):
             if not isinstance(text, str):
                 return np.nan
             if len(text) == 0:
-                return 0
-            sentences = sent_tokenize(text)
+                return 0.0
+            if not sentences := sent_tokenize(text):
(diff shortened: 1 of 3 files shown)


nlp_primitives/mean_characters_per_sentence.py
```diff
@@ -31,12 +31,9 @@ def get_function(self):
         def _mean_characters_per_sentence(text):
             if not isinstance(text, str):
                 return np.nan
-            if len(text) == 0:
-                return 0
-            sentences = sent_tokenize(text)
-            total = 0.0
-            for s in sentences:
-                total += len(s)
+            if not len(sentences := sent_tokenize(text)):
```
thehomebrewnerd, line 34: Might just be me, but I find this construct to take more cognitive energy to understand than:
```
sentences = sent_tokenize(text)
if not len(sentences):
    ...
```

Also it seems like it makes it a little more difficult to find where sentences was created since we can end up using sentences outside of the `if` block in which the assignment was created.

Not a blocker for me one way or the other, but I would tend to prefer an approach without the walrus operator here.

thehomebrewnerd (dismissed):

thehomebrewnerd (approved):

PR #251, opened 2023-03-07
Add OpenAI Embeddings Primitive

jlouns:
Adds a primitive for natural language logical types that uses the [OpenAI Embeddings API](https://platform.openai.com/docs/guides/embeddings) to calculate embeddings features.

The model to use is configurable, but `text-embedding-ada-002` is used by default.

### DIFF
--- a/nlp_primitives/openai/embeddings.py
+++ b/nlp_primitives/openai/embeddings.py
@@ -1,0 +1,190 @@
+import itertools
+from typing import List, Optional
+
+import numpy as np
+import pandas as pd
+import tiktoken
+from featuretools.primitives.base import TransformPrimitive
+from woodwork.column_schema import ColumnSchema
+from woodwork.logical_types import Double, NaturalLanguage
+
+from nlp_primitives.openai.api_requester import process_api_requests
+from nlp_primitives.openai.model import (
+    OpenAIEmbeddingModel,
+)
+from nlp_primitives.openai.request import (
+    OpenAIEmbeddingRequest,
+    OpenAIRequest,
+    StaticOpenAIEmbeddingRequest,
+)
+from nlp_primitives.openai.response import OpenAIEmbeddingResponse
+from nlp_primitives.utils import CurrentEventLoop
+
+DEFAULT_MODEL = OpenAIEmbeddingModel(
+    name="text-embedding-ada-002",
+    encoding="cl100k_base",
+    max_tokens=8191,
+    output_dimensions=1536,
+)
+
+
+class OpenAIEmbeddings(TransformPrimitive):
+    """Generates embeddings using OpenAI.
+
+    Description:
+        Given list of strings, determine the embeddings for each string, using
+        the OpenAI model.
+
+    Args:
+        model (OpenAIEmbeddingModel, optional): The model to use to produce embeddings.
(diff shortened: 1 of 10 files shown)


nlp_primitives/openai/embeddings.py
```diff
@@ -0,0 +1,186 @@
...
+        # mutable variables to track the request batching process
+        # a running list of the requests to make
+        requests: List[OpenAIRequest[OpenAIEmbeddingResponse]] = []
+        # a list of elements that should be batched into the next request
+        elements_in_batch = []
+        # a running total of tokens that will be sent in the next request
+        tokens_in_batch = 0
+
+        def add_batched_request() -> int:
+            """Create a batched request from the currently staged elements and add it to the request list
```
sbadithe, line 115:
```suggestion
            """Create a batched request from the currently staged elements and add it to the request list. Return the resulting total of tokens that will be sent in the next request. 
```

nlp_primitives/openai/embeddings.py
```diff
@@ -0,0 +1,186 @@
...
+                        token_consumption=tokens_in_batch,
+                    )
+                )
+                return 0
+            else:
+                return tokens_in_batch
+
+        def can_fit_in_batch(tokens) -> bool:
+            return (
+                len(elements_in_batch) < 2048
```
sbadithe, line 133: What is the significance of 2048?

nlp_primitives/openai/embeddings.py
```diff
@@ -0,0 +1,186 @@
...
+            if elements_in_batch:
+                elements_copy = list(elements_in_batch)
+                elements_in_batch.clear()
+                requests.append(
+                    OpenAIEmbeddingRequest(
+                        list_of_text=elements_copy,
+                        model=self.model,
+                        token_consumption=tokens_in_batch,
+                    )
+                )
```
sbadithe, line 126:
```suggestion
                requests.append(
                    OpenAIEmbeddingRequest(
                        list_of_text=elements_in_batch.copy(),
                        model=self.model,
                        token_consumption=tokens_in_batch,
                    )
                )
                elements_in_batch.clear()
```

Not sure if this is an improvement, feel free to disregard 😅

nlp_primitives/openai/embeddings.py
```diff
@@ -0,0 +1,186 @@
...
+                # add a static request that returns the invalid results
+                requests.append(invalid_request)
+            else:
+                # valid element
+                # check how many tokens are in it
+                next_tokens = len(encoding.encode(element))
+
+                # can this element fit in the batch?
+                if can_fit_in_batch(next_tokens):
+                    # can't fit -- construct a request with existing elements
```
sbadithe, line 152: Not sure if I am misunderstanding this, but should this block be the case where it can fit?

nlp_primitives/openai/embeddings.py
jlouns: yeah, here's the limit in the openai client: https://github.com/openai/openai-python/blob/main/openai/embeddings_utils.py#L43

nlp_primitives/openai/embeddings.py
jlouns: Good call. Fixed. I need to add tests for all of this to catch stuff like this 😅

PR #285, opened 2024-05-14
Support Python 3.12

### LINKED ISSUE #282: Add Python 3.12 support
Supporting Python 3.12 will require removing the tensorflow upper bound restrictions in #270 as only the latest version of tensorflow and tensorflow-hub support Python 3.12. However, these new tensor flow versions have removed deprecated functions that we are using in the Elmo and Universal Sentence Encoder primitives, so these primitives will need to be updated as well.

thehomebrewnerd:
Closes #282


thehomebrewnerd:
Will revisit in the future.

